Issue #10, opened 2023-04-22
Support resolving constant variable values


poteboy:
Currently, the `zero-styled` only supports inline values for style properties. It would be helpful if the plugin could also resolve constant variable values, both defined in the same file and imported from other files.

Example:

```jsx
const color = 'orange'
<Text fontSize="40px" color={color} />
```

poteboy:
The plugin now supports constant variable values defined in the same file, as addressed in PR #11. The support for resolving imported values from other files is still pending and will be addressed in a separate PR.

paulm17:
> The support for resolving imported values from other files is still pending and will be addressed in a separate PR.

Will this be available with the announced July 16th update?

poteboy:
Hey @paulm17 , thanks for the comment!
While the support for resolving imported values from other files is technically feasible at build time, it's worth noting that this approach could slow down HMR during development mode. To maintain an efficient DX, we will instead handle these values dynamically at runtime with our Hybrid approach.

However, we're considering doing static analysis for production builds specifically targeting imported values from other files in future updates. This feature might not be fully available in the July 16th update, but we're certainly moving towards that direction.

PR #11, opened 2023-04-22
Add support for resolving constant variable values

poteboy:
This PR adds support for resolving constant variable values in style properties #10


poteboy:
This PR addresses part of the issue #11  by allowing the plugin to resolve constant variable values defined in the same file. However, the support for resolving imported values from other files is still pending and will be addressed in a separate PR.

PR #24, opened 2023-05-04
Refactor HTML Tags.

yodakaEngineer:
Refactor htmlTags, and remove isHTMLElement method.

### DIFF
--- a/packages/core/src/k.ts
+++ b/packages/core/src/k.ts
@@ -1,14 +1,14 @@
 import { StyledComponentProps } from "./styled";
 import { StyledProps, StyledKeyType, PseudoProps } from "@kuma-ui/system";
 
-type StyledComponent<T extends keyof HTMLElementTagNameMap> = React.FC<
+type StyledComponent<T extends keyof JSX.IntrinsicElements> = React.FC<
   Omit<StyledComponentProps<T>, StyledKeyType> &
     StyledProps &
     Partial<PseudoProps>
 >;
 
 const k: {
-  [K in keyof HTMLElementTagNameMap]: StyledComponent<K>;
+  [K in keyof JSX.IntrinsicElements]: StyledComponent<K>;
 } = {} as any;
 
 export { k };
--- a/packages/core/src/k.ts
+++ b/packages/core/src/k.ts
@@ -33,11 +33,4 @@ const k: {
   [K in (typeof htmlTags)[number]]: StyledComponent<K>;
 } = {} as any;
 
-export const isHTMLElement = (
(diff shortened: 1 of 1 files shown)


packages/core/src/k.ts
```diff
@@ -1,43 +1,14 @@
...
-  [K in (typeof htmlTags)[number]]: StyledComponent<K>;
+  [K in (keyof HTMLElementTagNameMap)[number]]: StyledComponent<K>;
 } = {} as any;
 
-export const isHTMLElement = (
-  _tag: unknown
-): _tag is keyof JSX.IntrinsicElements => {
-  const tag = _tag as keyof JSX.IntrinsicElements;
-  return htmlTags.includes(tag);
-};
```
yodakaEngineer, line 41: This was not used.

packages/core/src/k.ts
```diff
@@ -1,43 +1,14 @@
 import { StyledComponentProps } from "./styled";
 import { StyledProps, StyledKeyType, PseudoProps } from "@kuma-ui/system";
 
-const htmlTags: Array<Partial<keyof JSX.IntrinsicElements>> = [
```
yodakaEngineer, line 4: Or, Maybe `keyof JSX.IntrinsicElements` is prefer instead of `HTMLElementTagNameMap`...?

packages/core/src/k.ts
```diff
@@ -1,43 +1,14 @@
...
-  "button",
-  "ul",
-  "ol",
-  "li",
-  "img",
-  "span",
-];
-
-type StyledComponent<T extends (typeof htmlTags)[number]> = React.FC<
+type StyledComponent<T extends keyof HTMLElementTagNameMap> = React.FC<
```
yodakaEngineer, line 4: Or, Maybe keyof JSX.IntrinsicElements is prefer instead of HTMLElementTagNameMap...?

yodakaEngineer: Or there is also keyof ReactHTML.
JSX.IntrinsicElements includes not HTML tag name.

poteboy: In my opinion, both `HTMLElementTagNameMap` and `ReactHTML` should be fine as long as they cover all HTML tags. However, in this particular case with Kuma UI, since the styled function's type is built upon keyof `JSX.IntrinsicElements`, it would be more appropriate to use `JSX.IntrinsicElements.` This ensures better consistency within the library and aligns with the existing type definitions 👍

poteboy (approved):
LGTM💜 Thank you so much!

PR #25, opened 2023-05-04
Fix incorrect setup example for Next.js

Karibash:
The gitkeep file was still there, so I removed it while I was at it. 1341ae0


poteboy (approved):
LGTM🍑

PR #29, opened 2023-05-05
Fix incorrect tsup config

Karibash:
There were a few things wrong with the tsup settings, which have been corrected.


poteboy:
@Karibash 
Thank you for the PR! I appreciate the effort you put into improving our build process. However, after pulling your changes and testing them locally, I encountered a new issue. When importing packages that depend on each other, I get the following error: `Could not find a declaration file for module '@kuma-ui/***'`, even though the declaration files are being generated correctly.

Can you please check if you can reproduce this issue in your local environment? If so, kindly include the fix in your pull request 🙏

Karibash:
@poteboy 
Hmmm, I can build and test locally without any problems.
GitHub Actions also pass without any problems, so I don't know why the error occurs in your environment.
Can you provide more detailed error messages or screenshots?

poteboy (approved):
Sorry it was my fault 🙏 I was using the TypeScript version from VSCode instead of the one specified in the workspace. I have added a small commit to ensure that every user can reference the same TypeScript version consistently.

poteboy:
It turns out that the issue was not related to the VSCode TypeScript version. The problem was resolved by changing the dependency resolution from `workspace:^` to `workspace:*` in the `packages/vite/package.json`.  https://github.com/poteboy/kuma-ui/commit/b1265ff47d980dc86d573f38010c459f17eee932

PR #36, opened 2023-05-07
Organize package dependencies

Karibash:
Removed unnecessary dependencies and relocated package specific dependencies from the root to the appropriate package.

### DIFF
--- a/configs/jest-preset/package.json
+++ b/configs/jest-preset/package.json
@@ -7,7 +7,8 @@
     "build": "tsup"
   },
   "dependencies": {
-    "ts-jest": "^29.1.0"
+    "ts-jest": "^29.1.0",
+    "ts-node": "^10.9.1"
   },
   "devDependencies": {
     "@jest/types": "^29.1.0"
--- a/package.json
+++ b/package.json
@@ -38,10 +38,7 @@
   },
   "devDependencies": {
     "@changesets/cli": "^2.26.1",
-    "@types/babel__core": "^7.20.0",
     "@types/node": "^18.15.11",
-    "@types/react": "^18.0.32",
-    "@types/webpack-sources": "^3.2.0",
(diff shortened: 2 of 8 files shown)


poteboy:
Thanks!
Kinda off-topic though, after applying the changes from your previous PR, I'm encountering an error with the `tsconfig.json` in each package. It seems to be caused by the addition of the `jest.config.ts` file at the same level as the `src` folder. The error message is as follows:
```sh
File '/path/to/jest.config.ts' is not under 'rootDir' '/path/to/src'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Matched by default include pattern '**/*'
```
One potential solution is to change the rootDir to '.', but I'm a bit hesitant to make this change. Would it be better to change the file extension of `jest.config.ts` to `.js` instead? I would like to hear your thoughts on how to proceed with this issue. @Karibash

Karibash:
@poteboy 
I could not see that error in my local environment.
When does the error occur? (during build or testing?)

poteboy:
@Karibash 
スクリーンショット 2023-05-07 20 41 25

Karibash:
@poteboy 
Ah, I understand now.
How about making it controlled by the include option as shown below instead of rootDir?

```json
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "paths": {},
    "types": ["node"],
    "declaration": true,
    "declarationDir": "dist"
  },
  "include": [
    "src/**/*.ts",
    "jest.config.ts"
  ]
}
```

poteboy:
@Karibash 
Good idea 👍 However, it's not going to work as expected because the TypeScript compiler will look for files relative to the rootDir. To make it work, we should update the include option like this:
```tsconfig.json
"include": [
    "src/**/*.ts",
    "../jest.config.ts"
  ]
```
I will include this change in this PR.

poteboy (approved):

Issue #37, opened 2023-05-07
Add support for Next.js 13.4 App Router and React Server Components (RSC)


poteboy:
Next.js 13.4 has been released, and it introduces the stable version of App Router, along with improved support for React Server Components (RSC). We would like to update Kuma UI to be compatible with these new features.

In our initial tests, the Babel Plugin's AST traversal appears to be working as expected. However, Kuma UI's current implementation of runtime injection of the head element in development mode is not compatible with RSC.

PR #40, opened 2023-05-08
Add support for Next.js 13.4 App Router and React Server Components (RSC)

poteboy:
#37


poteboy:
we'll open new one as this conflicts with the current implementation

PR #43, opened 2023-05-11
Change css injection method from runtime inject to virtualLoader file.

taishinaritomi:
I have changed the CSS injection method from runtime inject code to virtualLoader.

Note:
1. The previous implementation has not yet been removed as comments.
2. It is unstable as only basic verification has been performed.
3. More work is needed to support NextAppDir.
4. vite-plugin has not been changed yet. [vite virtual modules](https://vitejs.dev/guide/api-plugin.html#virtual-modules-convention)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist/
 coverage/
 node_modules/
 .vscode
+.kuma/
--- a/example/vite/src/App.tsx
+++ b/example/vite/src/App.tsx
@@ -7,7 +7,7 @@ function App() {
         hello
       </k.div>
       <Text color="blue">world</Text>
-      <k.button p="8px 12px" fontSize={16} disabled>
+      <k.button p="10px 12px" fontSize={16} disabled>
         button
       </k.button>
     </div>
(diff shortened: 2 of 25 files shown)


poteboy:
@taishinaritomi 
Thank you very much for your PR, I really appreciate it 🙏 I haven't yet had the chance to review all the code changes, so I kindly ask for your patience. Meanwhile, I do have a question. 

I suppose this change could potentially make it compatible with React Server Components (RSC) as this change modifies the CSS injection method from runtime inject code to using a virtual loader. However I am curious as to why these changes would not be compatible with App Router yet?

taishinaritomi:
In NextAppDir, the virtual loader does not work, so you have to actually write the css to the file (there may be a better way, but I am still working on it), and you also have to make some changes to next globalCssLoader.

https://github.com/taishinaritomi/kaze-style/blob/main/packages/webpack-plugin/src/transformLoader.ts#L94

https://github.com/taishinaritomi/kaze-style/blob/main/packages/webpack-plugin/src/transformLoader.ts#L94

poteboy:
this PR was mistakenly closed. I've incorporated the changes into a new PR, which is now open at [#44](https://github.com/poteboy/kuma-ui/pull/44)

PR #44, opened 2023-05-11
Change css injection method from runtime inject to virtualLoader file. 

poteboy:
This PR is a continuation of the work done in [#43](https://github.com/poteboy/kuma-ui/pull/43). Unfortunately, the original PR was mistakenly closed. I've reopened it here to incorporate the changes and continue the discussion.

PR #45, opened 2023-05-11
Add example directory to the workspace

poteboy:
This PR aims to include the example directory in the workspace managed by pnpm. Previously, projects under the example directory were unable to reference other packages in the workspace due to not being included in the pnpm-workspace.yaml file.

Interestingly, in the past, even though the example directory was not included in pnpm-workspace.yaml, dependencies were still being properly resolved when running `pnpm i` from the root directory. This allowed projects under the example directory to function correctly. However, after re-cloning the repository, this no longer seems to be the case. The reason for this change in behavior is currently unknown.

By adding the example directory into pnpm-workspace.yaml, we aim to ensure these projects can directly reference and use other packages within the workspace, regardless of the mysterious past behavior.

However, please note that currently, the build process for the example/next project is failing. As a temporary measure, we've renamed its build command to prevent this project from being built along with the rest of the workspace. We plan to address this build failure in a future update.

### DIFF
--- a/example/next/package.json
+++ b/example/next/package.json
@@ -9,8 +9,8 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@kuma-ui/core": "workspace:^",
-    "@kuma-ui/next-plugin": "workspace:^",
+    "@kuma-ui/core": "workspace:*",
+    "@kuma-ui/next-plugin": "workspace:*",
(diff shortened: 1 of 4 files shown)


poteboy:
This PR currently conflicts with PR #44. Therefore, we'll need to wait until PR #44 is merged before proceeding with this one.

PR #47, opened 2023-05-15
Support Next App Router & React Server Components

taishinaritomi:
@poteboy

This pull request adds support for Next App Router and React Server Components.

Bug:
There is a specification issue where it keeps creating CSS files in the "./.kuma" directory. Deleting them disrupts the dev build due to next cache.

Note:
This approach may not be the official method, and there might be better alternatives.
It is still unstable and requires validation.


poteboy:
First of all, thank you for your work on this pull request! I really appreciate it 🙏

I've tested it out and noticed that every time a component's style gets updated, a new CSS file is created and the old one remains. Is this an intentional behavior?

packages/webpack-plugin/src/loader.ts
```diff
@@ -61,7 +83,8 @@ export default kumaUiLoader;
 const requireReact = (code: string, id: string) => {
   if (id.endsWith(".jsx") || id.endsWith(".tsx")) {
     if (!/^\s*import\s+React\s+from\s+['"]react['"]/.test(code)) {
-      return "import React from 'react';\n" + code;
+      // return "import React from 'react';\n" + code;
+      return code;
```
poteboy, line 87: Could you explain the reason for commenting out this line and changing it to `return code;`? I may be misremembering, but I thought we had to import React because we're using 'classic' JSX runtime 👀

taishinaritomi: The reason for removing this is that the 'use clinet'; needs to be on the first line, and adding it would cause a crash. Also, I don't think adding 'react import' is a style library responsibility.

taishinaritomi:
> I've tested it out and noticed that every time a component's style gets updated, a new CSS file is created and the old one remains. Is this an intentional behavior? I suppose that it is because deleting the old CSS files immediately might cause issues if there are still references to them?

It is done intentionally, but it should not be done.
This is because next.js references the old CSS (we need to find a remedy).

poteboy:
Thank you for the clarification. I understand the situation now, and I agree that we'll need to find a solution for this in the future 🙏

I have a question regarding the compatibility between App Router and Pages Directory. Since Pages Directory does not support importing global CSS files into a component, would it break if we tried to combine App Router with Pages Directory? I would appreciate your insights on this.  @taishinaritomi

taishinaritomi:
"getGlobalCssLoader" added to "@kuma-ui/next-plugin" allows you to load CSS files ignoring the global CSS import constraint of Next.js. Therefore, pages and app should be able to coexist.

poteboy:
> "getGlobalCssLoader" added to "@kuma-ui/next-plugin" allows you to load CSS files ignoring the global CSS import constraint of Next.js. Therefore, pages and app should be able to coexist.

:kami: 🚀

poteboy (approved):
Thanks 🙌

PR #51, opened 2023-05-18
Add `css` Function

poteboy:
The css function allows users to define styles either as tagged template literals or as JavaScript objects, similar to libraries like Emotion.

```tsx
import { css } from '@kuma-ui/core';

const Component = () => (
  <div className={css`font-size: 24px; color: red;`} />
);

// or

const Component = () => (
  <div
    className={css({
      fontSize: '24px',
      color: 'red',
    })}
  />
);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,29 +14,7 @@
 🐻 Kuma UI is a utility-first, zero-runtime CSS-in-JS library that offers an outstanding developer experience and optimized performance.
 
 
-```tsx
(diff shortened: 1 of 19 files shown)


poteboy:
I've decided to prioritize the JavaScript objects pattern for its superior type safety and ease of static analysis.

taishinaritomi (approved):
reviewed 👍

PR #57, opened 2023-05-21
Fixed pnpm lock changing every time.

taishinaritomi:
Due to a pnpm version mismatch, it was broken, so we narrowed down the version to v8.

1. The specification of lock file is different between v8 and v7, so they are unified in v8.
2. Deleted lock file once and refreshed it.

I would like you to verify that it is not broken by local and Merge it.


poteboy (approved):
LGTM, Thanks!

PR #68, opened 2023-05-24
Add opacity CSS property support

masa5555:
CSS `opacity` is not supported


masa5555:
I have created the toCssUnitWithPx function to address the issue where the toCssUnit function always converts numbers to pixels. This allows for branching between numeric-only CSS and pixel-based CSS. As a result, tests for properties like opacity and z-index are now passing.

PR #78, opened 2023-05-28
test: style cache key 

sumiren:
This PR
- extracts cache key calculation algorithm to function.
- adds test for the function.


sumiren:
I apologize for leaving my pull request unattended. I would like to resubmit it when I have the time, but I don't want it to be in the way, so I will close it for now.

PR #92, opened 2023-06-04
Migrate from jest to vitest

you-5805:
## Issue
https://github.com/poteboy/kuma-ui/issues/90

## Overview
Migrate the test runner from jest to vitest

## TODO
- [x] migrate @kuma-ui/sheet a12258ac8097d6e005b53e90c1a12dec93b830cd
- [x] migrate @kuma-ui/system 16041971e41a2450846579702bcc96ba618215ed
- [x] migrate @kuma-ui/babel-plugin 6e568bce88f7190aef0c25aa6366ab59c98dcc3c
- [x] rm jest a1d97c2322d100c9f913ddaf871888b88479ca88
- [ ] extract vitest config to workspace and reuse it **(if needed)**

## Comment
- For I don't deeply understand about the implementation and the context of this issue, changes may have something wrong. If there are any oversights, please let me know.
- This task only took me around 20 minutes, so feel free to request the change.


packages/babel-plugin/vitest.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    globals: true,
```
you-5805, line 5: only in @kuma-ui/babel-plugin, `globals` is set to `true` for babel-plugin-tester behavior.

poteboy:
@you-5805 
I appreciate your efforts on this pull request!

While trying to run the changes locally, I encountered an error with the dependency `@ampproject/remapping@2.2.1` being invalid. I noticed that the lockfileVersion in `pnpm-lock.yml` was changed to '6.1'. Would it be possible to revert this back to '6.0'? 

Apart from that,  I noticed you mentioned 'extract vitest config to workspace and reuse it (if needed)' in your TODO list. If it's possible to centralize the vitest configuration for reuse, that would be really appreciated.

Once these changes are made, I believe your PR will be good to go. Thank you!

you-5805:
@poteboy 
Thank you for your review!

- fix lockfile version d71f683
- reuse shared config for vitest 6c32264
  I tried to define configShared as workspace module, but it seem doesn't work correctly on ESM 🤔
  So I put the `vitest.shared.ts` on the root like the official example.
  https://vitest.dev/guide/workspace.html#configuration

The Vercel CI seems to be failing, but I'm unable to check the details🙏
On local,  `pnpm i --frozen-lockfile`, `pnpm build`, and `pnpm test` are passed.

poteboy (approved):
@you-5805 
Thank you for your quick fixes!

No worries about the Vercel CI issue; it's due to an unrelated App Router cache problem that we're currently waiting on a Next.js fix for. 

Everything else looks great! I appreciate your help with improving our test setup :)

PR #93, opened 2023-06-05
throw an error when accessing any property of 'k' at runtime

kotarella1110:
fixes the issue where an error was not being thrown when accessing properties of the `k`.

As is:

スクリーンショット 2023-06-05 19 38 17

To be:

スクリーンショット 2023-06-05 19 36 25


poteboy (approved):
LGTM! Thanks :)

PR #99, opened 2023-06-06
Automate Changeset Release through CI

tomoyuki-hiranuma:
I tried to automate the changeset release process with CI.

 In order to use the automation, you seem to need to install [this bot](https://github.com/apps/changeset-bot) to this repository. And it might be necessary to give the bot permissions to write and create PRs  in github settings.

And, I need your help because I have not been able to test with this repository and have been testing with my own repository, so please let me know if there are any problems. I hope we can solve them together.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
           version: 8.5.1
 
       - name: install packages
-        run: pnpm install --frozen-lockfile
+        run: pnpm install
 
       - name: Run test
         run: pnpm test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,3 +30,6 @@ jobs:
 
       - name: Run test
         run: pnpm test
+
+      - name: Build packages
+        run: pnpm build
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
           version: 8.5.1
 
       - name: install packages
-        run: pnpm install
+        run: pnpm install --frozen-lockfile
 
       - name: Run test
         run: pnpm test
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -26,7 +26,10 @@ jobs:
           version: 8.5.1
 
       - name: install packages
-        run: pnpm install
+        run: pnpm install --frozen-lockfile
```
poteboy, line 29: I guess that `--frozen-lockfile` is set to true by default in a CI environment, so we don't need to specify it?
https://pnpm.io/cli/install

.github/workflows/release.yml
```diff
@@ -25,5 +25,11 @@ jobs:
...
 
-      - name: Build packages
-        run: pnpm build
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          publish: pnpm release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
poteboy, line 34: When I tried to register an environment variable with the name "GITHUB_TOKEN", I encountered an error message saying "Variable name cannot start with 'GITHUB_' prefix." Can you change this part to a different name?

.github/workflows/main.yml
tomoyuki-hiranuma: I see 👍 
It's new learning for me.
I fixed it!!

[abca3c3](https://github.com/poteboy/kuma-ui/pull/99/commits/abca3c3756308f1f42957bc496f8eda58aef4b01)

.github/workflows/release.yml
tomoyuki-hiranuma: Probably "GITHUB_TOKEN" does not need to be registered. It should be handled internally.
https://docs.github.com/ja/actions/security-guides/automatic-token-authentication

PR #102, opened 2023-06-09
test(babel-plugin): add k api test

kotarella1110:
I added tests for the `k` API, which didn't exist before.
This makes it easier to debug babel plugin and ensures a certain level of quality.

### DIFF
--- a/packages/babel-plugin/src/__test__/testUtils.ts
+++ b/packages/babel-plugin/src/__test__/testUtils.ts
@@ -1,4 +1,4 @@
-import { transformSync } from "@babel/core";
+import { BabelFileResult, transformSync } from "@babel/core";
 import plugin from "../";
 import { types, template } from "@babel/core";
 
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/src/__test__/testUtils.ts
```diff
@@ -17,10 +18,18 @@ export function babelTransform(
...
-    filename: "test.tsx",
+    filename,
+  });
+
+  if (result === null || result.code == null)
+    throw new Error(`Could not transform`);
+
+  result = transformSync(result.code, {
+    plugins: [plugin({ types, template })],
+    filename,
```
kotarella1110, line 29: In actual scenarios, I believe the Kuma UI Babel plugin is loaded after reading Babel presets like React. Therefore, I have also improved the tests accordingly.

poteboy (approved):
LGTM!

PR #104, opened 2023-06-09
feat(babel-plugin): support jsx runtime automatic

### LINKED ISSUE #50: Update JSX Runtime from 'classic' to 'automatic' in Babel Plugin
Kuma UI currently uses the 'classic' JSX Runtime in the Babel Plugin to [manipulate the AST](https://github.com/poteboy/kuma-ui/blob/main/packages/babel-plugin/src/extractStyleProps/fromJSX.ts). The 'classic' runtime transforms JSX tags as follows:
```tsx
// Before
import React from 'react';

function App() {
  return <h1>Hello World</h1>;
}

// After
import React from 'react';

function App() {
  return React.createElement('h1', null, 'Hello world');
}
```

As a result, CallExpression is used to manipulate the JSX AST, not JSXElement.

However, since React v17.0.0, the 'automatic' JSX runtime has become the standard as it offers a number of benefits such as improved bundle size and the ability to use JSX without importing React.

This issue aims to update the JSX runtime from 'classic' to 'automatic' in the Babel Plugin to be in line with this standard. This change would simplify the use of Kuma UI for users as they would no longer need to explicitly import React. Additionally, it would also potentially result in smaller bundle sizes. This change would require modifications to the AST manipulation process to work with JSXElement instead of CallExpression.

Reference: [Introducing the New JSX Transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)

kotarella1110:
## Resolve Issue

close #50 

## Overview

This pull request addresses the issue of **transforming from `React.createElement` to `React.createElement`, allowing for `JSX` to `JSX` transformation instead**. This change eliminates the dependency on React's runtime, making it possible to perform the transformation regardless of whether the runtime is `classic` or `automatic`. 
Additionally, switching from `JSX` to `JSX` transformation opens up the possibility of supporting UI libraries other than React (such as SolidJS) more easily.

## Details

The following outlines the transformation process before and after the changes made in this pull request:

Assuming the initial code is as follows:

```tsx
import { k } from '@kuma-ui/core'
function App() {
  return <k.div p={2} _after={{ color: 'blue' }} />
}
```

### Before

1. `JSX` is transformed to `React.createElement()` using presets like `@babel/plugin/preset-react`.

```js
function App() {
  return /*#__PURE__*/React.createElement(k.div, {
    p: 2,
    _after: {
      color: 'blue'
    }
  });
}
```

2. Subsequently, the Kuma UI plugin performs **`React.createElement` to `React.createElement`** transformation.

```js
.kuma-2631981251{padding:2px;}.kuma-2631981251:after{color:blue;}

import React from "react";
import { k } from '@kuma-ui/core';
function App() {
  return /*#__PURE__*/React.createElement("div", {
    className: ["kuma-2631981251"].join(" ")
  });
}
```

### After

1. The Kuma UI babel plugin performs **`JSX` to `JSX`** transformation.

```jsx
.kuma-2631981251{padding:2px;}.kuma-2631981251:after{color:blue;}

import React from "react";
import { k } from '@kuma-ui/core';
function App() {
  return <div className={["kuma-2631981251"].join(" ")} />;
}
```

2. The transformed JSX code from Kuma UI plugin is then transformed to `React.createElement()` or `jsx()` using plugins like `@babel/plugin/preset-react`, depending on the runtime configuration.

classic runtime:

```js
.kuma-2631981251{padding:2px;}.kuma-2631981251:after{color:blue;}

import React from "react";
function App() {
  return /*#__PURE__*/React.createElement("div", {
    className: ["kuma-2631981251"].join(" ")
  });
}
```

automatic runtime:

```js
.kuma-2631981251{padding:2px;}.kuma-2631981251:after{color:blue;}

import React from "react";
import { jsx as _jsx } from "react/jsx-runtime";
function App() {
  return /*#__PURE__*/_jsx("div", {
    className: ["kuma-2631981251"].join(" ")
  });
}
```

### DIFF
--- a/.changeset/heavy-trees-mate.md
+++ b/.changeset/heavy-trees-mate.md
@@ -1,0 +1,12 @@
+---
+"@kuma-ui/babel-plugin": minor
+"@kuma-ui/next-plugin": minor
+"@kuma-ui/vite": minor
+"next-app-router": patch
+"next": patch
+"react-vite-example": patch
+"webpack": patch
+"website": patch
+---
+
+feat(babel-plugin): support jsx runtime automatic


poteboy (approved):
Congrats 🚀 This is so good, nice work!

PR #115, opened 2023-06-16
chore: Automate Synchronization of README Files Across Packages

tkc310:
## Summary

Added script to sync `./README.md` to each `packages/README.md` at pre-commit to resolve the following issue.
https://github.com/poteboy/kuma-ui/issues/96

## Behavior

- Copy the contents of `./README.md` to `packages/README.md`.
- If there are no diffs, it is not executed.
- Modified the `lint-staged` section of package.json to run at pre-commit time.
**(Please run `pnpm husky install` after the merge to update your githook)**

## Unsupported

The problem is that `packages/README.md` and `./README.md` will be executed even when there is no diff.

However, since the timing of execution is at pre-commit and the differences remain on the stage after commit, no action was taken.
(We figured that if there were developers who only changed `package/README.md` they would notice at this time.)

## Comments

Please let me know if there are any errors due to my unfamiliarity 🙏

### DIFF
--- a/.changeset/add-sync-files.md
+++ b/.changeset/add-sync-files.md
@@ -1,0 +1,11 @@
+---
+"@kuma-ui/webpack-plugin": patch
+"@kuma-ui/babel-plugin": patch
+"@kuma-ui/next-plugin": patch
+"@kuma-ui/system": patch
+"@kuma-ui/sheet": patch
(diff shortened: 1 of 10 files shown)


tkc310:
Thanks for the prompt confirmation!

PR #116, opened 2023-06-16
fix(babel-plugin): fix issue with rewriting className attribute in processJSXHTMLTag function

kotarella1110:


packages/babel-plugin/src/processJSXHTMLTag.ts
```diff
@@ -29,43 +29,44 @@ export const processJSXHTMLTag = (path: NodePath<t.JSXOpeningElement>) => {
...
       ) {
-        if (t.isStringLiteral(attr.value)) {
-          classNameAttrs.push(attr.value);
-        } else if (
-          t.isJSXExpressionContainer(attr.value) &&
-          t.isExpression(attr.value.expression)
-        ) {
-          classNameAttrs.push(attr.value.expression);
-        }
-        return attr.value;
```
kotarella1110, line 50: This code is the cause of the Issue: whenever a className is specified in the JSX, it is returned here. Therefore, subsequent processing will not be executed.

packages/babel-plugin/src/__test__/k.test.ts
```diff
@@ -73,6 +73,26 @@ describe("k api", () => {
...
+              className={css({ boxShadow: '0 1px 2px 0 rgba(0, 0, 0, 0.05)' })}
+            />
+          )
+        }
+        `;
+        // Act
+        const result = babelTransform(inputCode, runtime);
+        // Assert
+        expect(getExpectSnapshot(result)).toMatchSnapshot();
+      });
```
kotarella1110, line 95: added test for the issue  👍

packages/babel-plugin/src/processJSXHTMLTag.ts
poteboy: I see 👍

poteboy (approved):
LGTM! Thank you 🙌

PR #123, opened 2023-06-16
Set of Zero-runtime UI Components

### LINKED ISSUE #113: Introducing Universal UI Components from @kuma-ui/core
Our ultimate goal is to develop a zero-runtime headless component library that adheres to users' design systems while prioritizing performance. We believe our library should function seamlessly with any JSX library, and we strive to fully realize this potential.

This proposal is for the evolution of our existing package, @kuma-ui/core, to become the central hub for all shared logic that doesn't rely on the specifics of any JSX syntax or library. Inside the core, we will house components tailored for different JSX libraries such as React and Solid.

Our strategy involves providing a jsxMode option in the plugin or config which dictates the 'mode' of JSX to be used (e.g., solid, react, etc.). Depending on the selected jsxMode, the import path for components inside the @kuma-ui/core will be transformed accordingly using Babel.

This allows us to keep all the logic inside a single core component library (@kuma-ui/core) that contains all our design system's logic, and then to dynamically adjust how these components behave to fit the idiomatic ways of building components in each respective library.

For the components, we will continue our strategy of extracting static props at build time for generating styles, while also supporting runtime behaviors. This allows us to maintain our zero-runtime performance promise as much as possible while providing a wide range of flexibility for users.

poteboy:
close #113 🍑

The main focus of this PR is to integrate our `@kuma-ui/babel-plugin` with the kit to ensure build-time style extraction.


poteboy:
Components like `k.div` will ultimately be transformed into the `Box` from `@kuma-ui/core`. Other components such as `Spacer` or `Flex` will internally return a `Box` as well.

This approach consolidates the processing of props that could not be statically extracted all within the `Box`. For instance, processing of component-specific props would be handled within Spacer, and only Kuma's Style Props would be passed to the Box.

Because of this, the `Box` component is necessary, and it will be initially merged under the experimental branch.

PR #140, opened 2023-06-25
feat(core): runtime handling for dynamic style props

kotarella1110:
close https://github.com/poteboy/kuma-ui/issues/132

## Overview

Kuma UI has traditionally adopted a zero-runtime approach by extracting and generating CSS based on static analysis during the build process. However, static analysis has its limitations and cannot accommodate the generation of dynamic styles. In this Pull Request, we analyze the styles that can be statically interpreted at build time and only extract styles that couldn't be statically interpreted at runtime. This means that Kuma UI is transitioning to a hybrid CSS-in-JS approach, which integrates the advantages of both zero-runtime CSS-in-JS and runtime CSS-in-JS.

### Performance

There are two ways to add styles to the document, and in the production environment (`NODE_ENV=production`), speedy mode is enabled for improved performance.

Development:

Styles are added to the style tag using the [document.createTextNode](https://developer.mozilla.org/docs/Web/API/Document/createTextNode) API. This approach has lower performance but does not have any drawbacks in terms of editing in the DevTools.

Production:

CSS "rules" are added to the browser backed stylesheet using the [sheet.insertRule](https://developer.mozilla.org/ja/docs/Web/API/CSSStyleSheet/insertRule) API. This approach has high performance but comes with the drawback that styles cannot be edited in the DevTools.

### Support Next.js App Router

If there are dynamic props present, we return Client Component (StaticBox), and if they don't exist, we return Server Component (DynamicBox) to achieve this.

### Support SSR

I have built it to support SSR. If you are using Kuma UI with Next.js, currently, you need to handle it on the user's side as follows. However, we provide `@kuma-ui/next-plugin`, so we believe that these configurations will no longer be necessary in the future.

#### Pages Router

`pages/_document.tsx`:

```tsx
import { createStyleRegistry, StyleRegistry } from "@kuma-ui/core";

export default class MyDocument extends Document {
  static async getInitialProps(ctx: DocumentContext) {
    const registry = createStyleRegistry();
    const originalRenderPage = ctx.renderPage;

    try {
      ctx.renderPage = () =>
        originalRenderPage({
          enhanceApp: (App) => (props) =>
            (
              <StyleRegistry registry={registry}>
                <App {...props} />
              </StyleRegistry>
            ),
        });

      const initialProps = await Document.getInitialProps(ctx);
      return {
        ...initialProps,
        styles: [initialProps.styles, registry.styles()],
      };
    } finally {
      registry.flush();
    }
  }

  render() {
    // ...
  }
}
```

#### App Router

`app/registry.tsx`:

```tsx
"use client";

import React, { useState } from "react";
import { useServerInsertedHTML } from "next/navigation";
import { StyleRegistry, createStyleRegistry } from "@kuma-ui/core";

export function KumaRegistry({ children }: { children: React.ReactNode }) {
  const [registry] = useState(() => createStyleRegistry());

  useServerInsertedHTML(() => {
    const styles = jsxStyleRegistry.styles();
    jsxStyleRegistry.flush();
    return <>{styles}</>;
  });

  return <StyleRegistry registry={registry}>{children}</StyleRegistry>;
}
```

`app/layout.tsx`:

```tsx
import { KumaRegistry } from "./registry"
 
export default function RootLayout({
  children,
}: {
  children: React.ReactNode
}) {
  return (
    <html>
      <body>
        <KumaRegistry>{children}</KumaRegistry>
      </body>
    </html>
  );
}
```

## Demo

Below is a demo of handling dynamic styles using Next.js's app router.

https://github.com/poteboy/kuma-ui/assets/12913947/33867c4c-0b29-4fcf-9e11-3ba620132733

You can see the actual behavior from the example.

https://github.com/poteboy/kuma-ui/tree/feat/runtime-handling/example


poteboy (approved):
LGTM 🚀

PR #145, opened 2023-06-25
Add outline-* CSS property

masa5555:


packages/system/src/outline.ts
```diff
@@ -0,0 +1,40 @@
+import { OutlineKeys } from "./keys";
+import { CSSProperties, ResponsiveStyle } from "./types";
+import { applyResponsiveStyles } from "./responsive";
+
+export type OutlineProps = Partial<
+  CSSProperties<"outline"> &
+    CSSProperties<"outlineOffset"> &
+    CSSProperties<"outlineWidth"> &
```
poteboy, line 8:
```suggestion
    CSSProperties<"outlineWidth", true> &
```

`outlineWidth` should be able to accept number value

packages/system/src/outline.ts
```diff
@@ -0,0 +1,40 @@
...
+  const media: ResponsiveStyle["media"] = {};
+
+  for (const key in outlineMappings) {
+    const cssValue = props[key as OutlineKeys];
+    if (cssValue) {
+      const property = outlineMappings[key as OutlineKeys];
+      const responsiveStyles = applyResponsiveStyles(
+        property,
+        cssValue,
+        undefined
```
poteboy, line 30: `outlineWidth` should be transformed via `toCSSUnit` when its type is number

poteboy (approved):
LGTM! thanks 🐻‍❄️

PR #165, opened 2023-07-02
feat: support using theme's colors and breakpoints at runtime

kotarella1110:
## Summary

This PR improves the functionality to support dynamic style props even when a theme is included.

In this PR, we have added support for only `colors` and `breakpoints` within the theme, while the support for component `variants` is not included. Support for `variants` will do in a separate PR.

## Changes

The main changes include enabling the use of the `@kuma-ui/sheet`'s `Theme` class at runtime.

Details:

- To reduce the possibility of global variable conflicts, the naming has been changed from `KUMA_RUNTIME_USER_THEME` to `__KUMA_RUNTIME_USER_THEME__`.
- To utilize the userTheme at runtime, `__KUMA_USER_THEME__` has been added to `globalThis`.
- The `globalThis.__KUMA_USER_THEME__` and `globalThis.__KUMA_RUNTIME_USER_THEME__` are now used as the initial values for the theme class. This allows the theme class to be used at runtime.
- Removed `runtimeTheme` from `@kuma-ui/core` because we can use `@kuma-ui/sheet`'s `theme` at runtime.

## Related PR

https://github.com/poteboy/kuma-ui/pull/162


packages/core/src/runtimeTheme.ts
```diff
@@ -1,23 +0,0 @@
...
-}
-
-export function getTokens(): Record<string, string> {
-  return theme.tokens || {};
-}
-
-
-export function getBreakPoints(): Record<string, string> {
-  return theme.breakpoints || {};
-}
```
kotarella1110, line 23: I have removed these functions. Please use the theme methods as follows going forward.

```
import { theme } from "@kuma-ui/sheet";

theme.getValiants("Box");
theme.getTokens();
theme.getBreakPoints();
```

poteboy (approved):

taishinaritomi (approved):

PR #170, opened 2023-07-04
Implement caching for StyleGenerator usage in extractPropsCache props

poteboy:
In this PR, we have introduced caching to the extractProps function to optimize the usage of StyleGenerator, which is computationally heavy. By caching the result based on the props, we avoid unnecessary computations, thereby improving the performance.


kotarella1110 (approved):
awesome PR! 🎉

PR #174, opened 2023-07-05
fix(core): fix speedy mode issue

kotarella1110:
In order to improve the runtime performance in the production environment, speedy mode is enabled. When speedy mode is enabled, styles are updated using the CSSStyleSheet's insertRule API. However, we discovered an issue where the insertRule API throws an error when passed dynamic props like media queries to Kuma UI component.

https://github.com/poteboy/kuma-ui/assets/12913947/5f30a8e5-678e-4b8a-9932-1888c1f49168

This problem occurs because the insertRule API only accepts one rule at a time, as stated in the following documentation:　https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule

> Thrown if more than one rule is given in the rule parameter.

In practice, if multiple rules are passed, an error occurs, as shown in the example below:

```
document.styleSheets[0].insertRule(".kuma-2804477523{margin-bottom:6px;}@media (min-width:1000px){.kuma-2804477523{margin-bottom:12px;}}")
```

Therefore, it is necessary to pass each rule one by one, as demonstrated in the example below:

```
document.styleSheets[0].insertRule(".kuma-2804477523{margin-bottom:6px;}")
document.styleSheets[0].insertRule("@media (min-width:1000px){.kuma-2804477523{margin-bottom:12px;}}")
```

This pull request aims to resolve this issue.

### DIFF
--- a/packages/core/src/registry/StyleSheetRegistry.ts
+++ b/packages/core/src/registry/StyleSheetRegistry.ts
@@ -1,4 +1,5 @@
 import React from "react";
+import { compile, serialize, stringify } from "stylis";
 import { StyleSheet } from "./sheet/StyleSheet";
 import { isBrowser } from "../utils/isBrowser";
 import { isProduction } from "../utils/isProduction";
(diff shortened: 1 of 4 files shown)


packages/core/src/registry/StyleSheetRegistry.ts
```diff
@@ -7,93 +8,105 @@ const STYLE_ID_PREFIX = "__";
...
-      this.indices[id] = this.sheet.insertRule(rule, this.indices[id]);
+    this.instancesCountMap[id] = 1 + (this.instancesCountMap[id] ?? 0);
+    const serverSideRenderedStyle = this.serverSideRenderedStyleMap?.[id];
+    if (this.instancesCountMap[id] === 1 && !serverSideRenderedStyle) {
+      compile(css).forEach((element) => {
+        const rule = serialize([element], stringify);
+        this.indexesMap[id] = (this.indexesMap[id] || []).concat(
+          this.sheet.insertRule(rule)
+        );
+      });
```
kotarella1110, line 38: fixed to process multiple rule strings using `stylis` and insert each rule one by one.

packages/core/src/registry/StyleSheetRegistry.ts
```diff
@@ -7,93 +8,105 @@ const STYLE_ID_PREFIX = "__";
...
           key: `${STYLE_ID_PREFIX}${id}`,
           nonce: options.nonce ? options.nonce : undefined,
           dangerouslySetInnerHTML: {
-            __html: rule.cssText,
+            __html: css,
           },
         });
       })
-      .filter((props) => props !== null);
+      .filter(Boolean);
```
kotarella1110, line 98: added code to concatenate the CSS rules obtained from the indexes because fixed to have multiple indexes per id (className).

poteboy (approved):

PR #178, opened 2023-07-06
IS_KUMA_DEFAULT

poteboy:
The `Box` component acts as a base for all other components in Kuma UI. However, we found that applying the `Box` variant to other components was causing unintended side effects. To counter this, we have now implemented a check that prevents the `Box` variant from being applied to other components. This ensures that the styles applied are those intended specifically for each individual component.

-  **Addition of IS_KUMA_DEFAULT Prop During Build Time:** 
During the build process, we now add a new prop, `IS_KUMA_DEFAULT`, to components created from the `styled` or `k` API. This prop is instrumental in determining whether a component should receive the default variant styles.

-  **Passing of IS_KUMA_DEFAULT Prop to Box at Runtime:** 
At runtime, the `IS_KUMA_DEFAULT` prop is passed from all components to the `Box` component. This enables us to determine whether the `Box` component should use its default variant styles or the styles of the component using it as a base.


kotarella1110 (approved):
LGTM 👍

kotarella1110 (approved):

PR #211, opened 2023-07-12
feat: support Waku

kotarella1110:


packages/core/src/components/componentList.ts
```diff
@@ -1,18 +1,3 @@
...
-  Text,
-  Heading,
-  Select,
-  Input,
-  HStack,
-  VStack,
-  Image,
-  Link,
-  Grid,
-} from "./index";
```
kotarella1110, line 15: "waku" is a React application running in a react-server environment with the vite plugin loaded.As a result, the entry point called "react.shared-subset.js" is loaded into React.

https://github.com/facebook/react/blob/0a360642dcc08cab469da0a0e4f1050e56b0ef7e/packages/react/package.json#L24

Therefore, `React` does not include `useState` or `createContext`.

https://github.com/facebook/react/blob/0a360642dcc08cab469da0a0e4f1050e56b0ef7e/packages/react/src/ReactSharedSubset.js

Importing a file that executes `createContext` would cause waku to malfunction, so I have made modifications to prevent that file from being loaded by vite. As a result, the dev server will now start up.

PR #212, opened 2023-07-13
chore: add codesandbox ci

kotarella1110:
ref: https://github.com/poteboy/kuma-ui/issues/208#issuecomment-1633409968


.codesandbox/ci.json
```diff
@@ -0,0 +1,7 @@
+{
+  "installCommand": "-v & pnpm i -g pnpm & pnpm i",
+  "buildCommand": "-v & pnpm build",
```
kotarella1110, line 3: `installCommand` and `build` are essentially unnecessary, but the following error will occur if you do not install the latest pnpm in advance.

https://github.com/poteboy/kuma-ui/blob/63a6f80e1eb4e6f36d871cedd7ff1c986dfa01c8/.codesandbox/ci.json

スクリーンショット 2023-07-13 20 38 21

I am using a somewhat unconventional method because when using `installCommand`, the command is executed using npm instead of pnpm, for some reason.

kotarella1110:
🎉
スクリーンショット 2023-07-13 20 37 19

poteboy (approved):
🙏🙏

PR #222, opened 2023-07-15
refactor: remove unnecessary processes

nojiritakeshi:
Deleted processing considered unnecessary in compiler/src/compile.ts.
I think the deleted processing is already compensated for by the processing in the find method.

Sorry for the trivial fix, but please check it out!


poteboy (approved):
LGTM :)

PR #224, opened 2023-07-16
fix: resolving paths on Windows

rabitarochan:
Fixed a problem where file paths could not be resolved in Windows.

With this change, we are also adding a CI build for Windows.

See memo (Japanese) : https://gist.github.com/rabitarochan/31ef1536209704923b57190983b9f3ee


poteboy:
I noticed that the original script (`packages/core/scripts/add-mjs-ext.mts`) was not working properly on a Windows environment. I suspect this might be due to differences in how file paths are structured on Unix and Windows systems. In the script, there are sections where regular expressions are used to process paths, and it appears these might not correctly handle Windows-style paths. Could you please confirm if my understanding is correct? @rabitarochan

rabitarochan:
@poteboy 
Yes, it is correct.

I have changed that fix from a script to the esbuild plugin. The reason for this is to incorporate it into the build process from the post-build patch script.

poteboy (approved):
Thank you so much!

PR #233, opened 2023-07-20
feat(compiler): static branching to handle conditional expressions

naruaway:
This is a continuation of https://github.com/poteboy/kuma-ui/pull/229

After discussing with core maintainers, we decided to separate class names per conditionals. 
In this way we no longer have to limit the combination limit up to arbitrary number such as "8" as seen in the previous PR.

The strong assumption here is that each style prop is isolated with each other. It's like "atomic CSS" but this PR just generates minimum number of classnames rather than actually generating separate classnames per each style prop.

Most of the things in [the previous PR description](https://github.com/poteboy/kuma-ui/pull/229) still applies. For example, if we decide to ship this PR, we should update documentation to clarify "static vs. dynamic"


packages/compiler/src/collector/collect.ts
```diff
@@ -10,15 +8,20 @@ import { match } from "ts-pattern";
...
   node: JsxOpeningElement | JsxSelfClosingElement
-) => {
+): Record<string, any> => {
   const jsxAttributes = node.getAttributes();
   const extracted: Record<string, any> = {};
+
   jsxAttributes.forEach((jsxAttribute) => {
     if (Node.isJsxAttribute(jsxAttribute)) {
-      const propName = jsxAttribute.getNameNode().getFullText();
+      const propName = jsxAttribute.getNameNode().getText();
```
naruaway, line 24: Note: @poteboy I guess the usage of `getFullText` was probably a mistake. Due to this we have several `trim()` in compiler package. I did not include the removals of `trim()` in this PR but we can remove them as well in later PRs

poteboy: Ah I see, thaks this helps a lot :)

naruaway:
Note on code structure: things like classifyProps lives in existing file rather than `static-branching.ts`, I might try refactor the structure more later although it should not affect the behavior

kotarella1110:
@naruaway please resolve conflict :pray:

naruaway:
@kotarella1110 thanks! Conflict resolved now
I guess additional processing such as pseudo props handling can be added later to reduce the risk of introducing many new bugs

Note that as discussed, one of potential bugs of this PR could be "style override rule can change between static and dynamic" but we think this PR does not make the problem worse since if any of our style props have conflicting output style, this problem already exists

packages/compiler/src/static-branching.ts
```diff
@@ -0,0 +1,21 @@
+export type AttributeValue =
```
poteboy, line 1: Since `AttributeValue` has already been defined in `compile.ts`, it seems redundant to me to redefine the same type.

packages/compiler/src/collector/collect.ts
```diff
@@ -50,6 +63,28 @@ const extractAttribute = (jsxAttribute: JsxAttribute) => {
         const expression = initializer.getExpression();
         if (!expression) return;
 
+        // fontSize={... ? ... : ...}
+        const conditionalExpression = match(expression)
+          .when(Node.isConditionalExpression, (conditional) => {
```
poteboy, line 68: I actually prefer this logic to be placed in `expression.ts` because, as a reader, it's more intuitive to have all expressions handled in `expression.ts`

naruaway: Sure, the initial reason why I put it here is to limit the scope of the change; but at this point probably better to clean and unify the logic there. Let me try a little bit refactoring although it will affect more such as pseduo props handling, which might be actually desired as long as.... it works without unexpected bug :)

naruaway: @poteboy, now I updated this PR to include some refactorings and pseudo props should be also handled properly for conditional expression (cc @kotarella1110) 👍

packages/babel-plugin/src/__test__/k.test.ts
```diff
@@ -74,6 +74,34 @@ describe("k api", () => {
...
+        // Assert
+        expect(getExpectSnapshot(result)).toMatchSnapshot();
+      });
+
+      test("using pseudo elements with top-level conditional should match snapshot", () => {
+        // Arrange
+        const inputCode = `
+        import { k } from '@kuma-ui/core'
+        function App({ flag }) {
+          return <k.div p={2} m={flag ? 100 : 10} _hover={flag ? { color: 'blue', bgColor: 'gray'} : {color: 'green'}} />
```
naruaway, line 96: Note: See the snapshot for how this is compiled. Note that conditional expression can appear at the top level of pseudo props or per prop level. If conditional expressions are nested, it will bail out

packages/compiler/src/collector/collect.ts
```diff
@@ -1,62 +1,54 @@
...
-  const extracted: Record<string, any> = {};
+  const extracted: Record<string, types.Value> = {};
+
   jsxAttributes.forEach((jsxAttribute) => {
     if (Node.isJsxAttribute(jsxAttribute)) {
-      const propName = jsxAttribute.getNameNode().getFullText();
-      let propValue;
-      // If the propName starts with underscore, use extractPseudoAttribute
-      if (propName.trim().startsWith("_")) {
-        propValue = extractPseudoAttribute(jsxAttribute);
```
naruaway, line 26: Note: now `extractPseudoAttribute` is gone and we just use the same function, `extractAttribute` for both

packages/compiler/src/collector/expression.ts
```diff
@@ -1,49 +1,124 @@
...
+  if (Node.isTypeAssertion(node)) {
+    return normalizeNode(node.getExpression());
+  }
+
+  return node;
+};
+
+export const handleJsxExpression = (
+  node: Node<ts.Node>
+): types.Value | undefined => {
```
naruaway, line 30: Note: Now the semantics of the return value of this function is clearer: `undefined` means "failed to evaluate" and `types.Value` can contain `undefined`. Previously these are not separated and `<Box color={undefined} />` was not statically compiled since `undefined` means "it failed to extract in the logic"

packages/compiler/src/collector/expression.ts
```diff
@@ -1,49 +1,124 @@
 import { Node, ts } from "ts-morph";
 import { match } from "ts-pattern";
+import * as types from "../types";
 
-export const handleJsxExpression = (node: Node<ts.Node>) => {
+/**
+ * Normalize ts-morph Node to make further processing simpler
+ */
+const normalizeNode = (node: Node): Node => {
```
naruaway, line 8: Note: I copied this from `decode.ts`, which is now removed. Now this is used internally by `handleJsxExpression` consistently. Previous pattern was error prone since people might forget to call `decode` where necessary. I think there is no reason to split the call between `decode` and `normalizeNode`. Note that in the future probably we can replace this kind of logic with some library such as ts-evaluate

packages/compiler/src/collector/expression.ts
```diff
@@ -1,49 +1,124 @@
...
+              whenFalse: whenFalse.value,
+            })
+            : undefined;
+        return ret;
+      })
+      // _hover={{color: 'red'}}
       .when(Node.isObjectLiteralExpression, (obj) => {
-        // TODO
-        return undefined;
+        const entries: [string, types.Value][] = [];
```
naruaway, line 91: I mostly copied the logic in `pseudo.ts`, which is now removed.
Note that the original logic had the following bug, which is fixed in my PR:

## Computed properties were silently erased
`_hover` prop was removed from `<Box _hover={{ ['color']: 'red'}} />` without generating the red color CSS. In my PR, computed properties just deopt.

.github/workflows/main.yml
```diff
@@ -56,4 +56,4 @@ jobs:
         run: pnpm install
 
       - name: Run lint
-        run: pnpm lint
+        run: pnpm build && pnpm lint
```
naruaway, line 59: Note: I realized ESLint emits false positive errors without `pnpm build`. Probably the correct way to fix it to somehow create a command using turborepo for `lint` to depend on `build` but I am not familiar. This should work for now at least for PR pipelines

packages/compiler/src/extractor/extract.ts
```diff
@@ -151,25 +102,190 @@ export const extractProps = (
...
+    ].join(" "),
+  };
 };
 
 /**
  * Generates a unique key for props, aiding cache efficiency.
  * Incurs O(n log n) cost due to sorting, but it's acceptable given the
  * expensive nature of StyleGenerator's internals.
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any -- FIXME
```
naruaway, line 166: This ESLint error is an actual bug. This shows that proper TypeScript typing (basically, eliminating `any`) with type-aware ESLint rules are really powerful to detect bugs 🎉 

The error is the following:
```
error  Invalid type "StaticValue" of template literal expression  @typescript-eslint/restrict-template-expressions
```

This indicates that we are trying to serialize object into string, which is not expected. I tested `main` (`2648f5eb5ac8d8cb1a50dd8c84539e314eb4afaa`) and confirmed that [generateKey](https://github.com/poteboy/kuma-ui/blob/2648f5eb5ac8d8cb1a50dd8c84539e314eb4afaa/packages/compiler/src/extractor/extract.ts#L104-L104) returns something like `_hover:[object Object]|bg:#576ddf|borderRadius:14px|color:white|cursor:pointer|fontWeight:600|p:16px 32px`, which is broken since nested object is stringified as `[object Object]`.

naruaway: I filed https://github.com/poteboy/kuma-ui/issues/246

packages/compiler/src/extractor/extract.ts
```diff
@@ -151,25 +102,190 @@ export const extractProps = (
...
-    .filter(([, value]) => value !== undefined)
-    .sort((a, b) => a[0].localeCompare(b[0]))
-    .map(([key, value]) => `${key}:${value}`)
-    .join("|");
+const generateKey = (props: Record<string, types.StaticValue>) => {
+  return (
+    Object.entries(props)
+      .filter(([, value]) => value !== undefined)
+      .sort((a, b) => a[0].localeCompare(b[0]))
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- FIXME
```
naruaway, line 285: This ESLint error is an actual bug. This shows that proper TypeScript typing (basically, eliminating `any`) with type-aware ESLint rules are really powerful to detect bugs 🎉 

The error is the following:
```
error  Invalid type "StaticValue" of template literal expression  @typescript-eslint/restrict-template-expressions
```

This indicates that we are trying to serialize object into string, which is not expected. I tested `main` (`2648f5eb5ac8d8cb1a50dd8c84539e314eb4afaa`) and confirmed that [generateKey](https://github.com/poteboy/kuma-ui/blob/2648f5eb5ac8d8cb1a50dd8c84539e314eb4afaa/packages/compiler/src/extractor/extract.ts#L104-L104) returns something like `_hover:[object Object]|bg:#576ddf|borderRadius:14px|color:white|cursor:pointer|fontWeight:600|p:16px 32px`, which is broken since nested object is stringified as `[object Object]`.

naruaway: I filed https://github.com/poteboy/kuma-ui/issues/246

packages/compiler/src/types.ts
```diff
@@ -0,0 +1,45 @@
...
+    };
+
+export type Conditional = {
+  type: "Conditional";
+  expression: string;
+  whenTrue: StaticValue;
+  whenFalse: StaticValue;
+};
+
+export type Value =
```
naruaway, line 18: Note: This is the most important type in this PR, which tries to encode the potential patterns (e.g. we see that conditional can only contain `StaticValue`, which means that nested conditionals are not allowed)

packages/compiler/src/collector/collect.ts
poteboy: With the current changes, the logic for determining the pseudo prop has been removed. This is unsafe, so it is necessary to add similar determination logic in `handleJsxExpression` by tracing back to the parent node from the expression.

packages/compiler/src/collector/expression.ts
```diff
@@ -1,49 +1,122 @@
...
+          })
+          : undefined;
+      })
+      // _hover={{color: 'red'}}
       .when(Node.isObjectLiteralExpression, (obj) => {
-        // TODO
-        return undefined;
+        const entries: [string, types.Value][] = [];
+        for (const prop of obj.getProperties()) {
+          if (Node.isPropertyAssignment(prop)) {
```
poteboy, line 91: I prefer early returns over full-body conditional wrapping in function declaration :)

packages/compiler/src/collector/expression.ts
```diff
@@ -1,49 +1,122 @@
...
+          } else {
+            return undefined;
+          }
+        }
+
+        if (entries.every((e) => e[1].type === "Static")) {
+          return types.staticValue(
+            Object.fromEntries(
+              entries.map((e) =>
+                e[1].type === "Static" ? [e[0], e[1].value] : ({} as never)
```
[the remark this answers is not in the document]
naruaway: Oh right, actually this is accidental since the argument of staticValue is `StaticValue` but  the type of `Object.fromEntries(entries)` is `{ [k: string]: types.Value; }`. Ideally type check should fail and somehow I thought it fails but it actually passes 🤦 

Anyway the runtime logic is correct so as long as we cover this part in test cases it should be acceptable, thanks for removing redundant checks 👍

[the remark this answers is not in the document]
naruaway: I also missed this although I wrote it: it's because `e[1].value` is needed to unwrap the value from `Static` variant. `e[1]` is like `{type: "Static", value: /* actual value here */ }`

PR #237, opened 2023-07-21
fix: correctly compile falsy property values

mtripg6666tdr:
(This original post has been edited to reflect correctly the current content of PR.)
kuma-ui has some utility props but if the value of the properties are zero, the properties will be removed in spite of it is necessary.
Currently this behavior can be dismissed by the workaround of using a string `"0"` instead of numerical `0`.
This PR makes the compiler recognize falsy values except `null` and `undefined`, as valid value and fixes the behavior.

<details>
<summary>Original Post</summary>
kuma-ui has some utility props that can change components' spaces, such as m and p but if the value of the properties are zero, the properties will be removed in spite of it is necessary.
Currently this behavior can be dismissed by the workaround of using a string "0" instead of numerical 0.
This PR makes the compiler recognize numerical 0 as valid value and fixes the behavior.
</details>

### DIFF
--- a/packages/compiler/src/collector/collect.ts
+++ b/packages/compiler/src/collector/collect.ts
@@ -27,7 +27,7 @@ export const collectPropsFromJsx = (
         propValue = extractAttribute(jsxAttribute);
       }
       // If the value is returned, it means that it can be statically analyzed, so we remove the corresponding prop from the Jsx tag and generate CSS.
-      if (!propValue) return;
+      if (!propValue && propValue !== 0) return;
       extracted[propName] = propValue;
     }
   });
--- a/packages/compiler/src/collector/collect.ts
+++ b/packages/compiler/src/collector/collect.ts
@@ -27,7 +27,7 @@ export const collectPropsFromJsx = (
         propValue = extractAttribute(jsxAttribute);
       }
       // If the value is returned, it means that it can be statically analyzed, so we remove the corresponding prop from the Jsx tag and generate CSS.
-      if (!propValue && propValue !== 0) return;
+      if (propValue == undefined) return;
       extracted[propName] = propValue;
     }
   });
--- a/packages/system/src/animation.test.ts
+++ b/packages/system/src/animation.test.ts
@@ -22,6 +22,7 @@ describe("animation utility function", () => {
       "animation-timing-function: ease-in;",
       "",
     ],
+    [{ animationIterationCount: 0 }, "animation-iteration-count: 0;", ""],
   ];
 
   test.each(testCases)(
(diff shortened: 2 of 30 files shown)


poteboy:
@mtripg6666tdr 
Thank you! If you're up for it, would you mind adding this processing to the other utilities within the `system` as well?

poteboy:
By the way, using `!propValue` not only removes 0 but also empty strings. However, there may be cases where you want to insert an empty string into the content of `::before` or `::after`. I thought it might be good to create and reuse a function like `isNonEmpty`. What do you guys think? @kotarella1110  @taishinaritomi

mtripg6666tdr:
Sure, it's good idea. Probably in many cases, what should be absolutely purged are `null` and `undefined`.
Considering empty strings are in many cases used as value of `content`, we may have to purge CSSs by seeing its property names and value, not only its value.

Anyway, if we use some so complex process that it can't write as one javascript statement, it's good to extract the utility method.

naruaway:
(Not directly relevant to the main point, but FYI "the value" of `content` never becomes "empty string" in JS representation since empty string is represented as quoted string like `content: "";` (CSS) / `{content: '""'}` (JSON)). This is because we can also have non string values such as `none` `content: none;` (CSS) / `{content: 'none'}` (JSON). cf. [MDN doc](https://developer.mozilla.org/en-US/docs/Web/CSS/content))

poteboy:
@naruaway 
Thanks for the clarification! Considering this, how do you envision the `isNonEmpty` function should be implemented?

naruaway:
I think actually there is no need to treat `0` in a special way. So there is also no need to do anything about `content` property.

"Emptiness" is already expressed as `undefined`. I feel like something like `propValue === undefined` (and `propValue !== undefined`) is more correct

kotarella1110:
I agree with @naruaway , I checked the TS type and it seems that you just need to exclude undefined. if you want to exclude `null` and `undefined` in consideration of JS users, you can use the equality (`==`) operator (`propValue == undefined`).

mtripg6666tdr:
So... `propValue != undefined` is most appropriate to be a judgment way of excluded properties, right? If so, I'll apply this change to the other properties not only "space"s. Where should I put the utility method?

poteboy:
@mtripg6666tdr 
Yes, thank you. I believe If the function only consists of the '==' operator, we might not need to separate it out into a function.

mtripg6666tdr:
I just updated changes to reflect the discussion.

poteboy:
Thanks! Could you add some test cases?

mtripg6666tdr:
Should I add test cases of property values of null or undefined?

poteboy:
would be sufficient if we could ensure that a numerical 0 is properly converted 👍

mtripg6666tdr:
then i think all set!

poteboy (approved):
LGTM 🙌

PR #238, opened 2023-07-21
chore: Introduce @typescript-eslint/recommended-type-checked and update ESLint packages

naruaway:
`@typescript-eslint/recommended-type-checked` is very powerful. See [the official doc](https://typescript-eslint.io/linting/typed-linting/) for its tradeoffs. Basically ESLint will be slower since it will use type information. But IMHO it's worth it.

For example, https://github.com/poteboy/kuma-ui/pull/235 could have been pointed out by [await-thenable](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/await-thenable.md) rule.

However, this introduced too many new ESLint errors. I do not have a capacity to fix all of them in a single PR.
So I just introduced and ran [eslint-interactive](https://github.com/mizdra/eslint-interactive) to add "eslint-disable" comments. We can gradually fix them. IMHO having several FIXME comments is better than not having ESLint rules.


packages/babel-plugin/src/transform.ts
```diff
@@ -3,6 +3,7 @@ import { compile } from "@kuma-ui/compiler";
 import pluin from ".";
 import { sheet } from "@kuma-ui/sheet";
 
+// eslint-disable-next-line @typescript-eslint/require-await -- FIXME
```
naruaway, line 6: Example explanation: now this function does not include any `await` so ESLint is pointing out that :)

poteboy:
@naruaway 
Thanks for the PR!
Can you disable the `@typescript-eslint/no-unused-vars rule`? I find it a bit bothersome when modules I've imported with the intention of using later get removed while I'm still editing the file. Unless I'm mistaken, bundlers like `esbuild` automatically delete unused variables, so I'm assuming it doesn't affect performance?

naruaway:
@poteboy Sure, let me remove that 👍 
Just to clarify, I think the primary benefit of using this rule is to encourage not having unused part of code to avoid confusion for people who is reading code. Also it can detect forgotten original intention such as [this `pluginName` unused variable](https://github.com/poteboy/kuma-ui/blob/0a911a86f82d50b5fa8c0fd6abf61d4df25d3a1c/packages/webpack-plugin/src/plugin.ts#L16-L16).

But I understand the frustration for that workflow. I personally sometimes turn off ESLint realtime checker while editing files since there are also other rules (and even TypeScript checks) which is noisy while quickly prototyping things

For performance, I think you are right, most of the time, unused variables should be just dropped from the final output. However, if the target expression is slightly more complicated than simple constant assignment, tools like esbuild cannot figure out actual unused code and it will remain in the final bundle like [this example](https://esbuild.egoist.dev/#W1siaW5kZXgudHMiLHsiY29udGVudCI6ImltcG9ydCB7IHN1bSB9IGZyb20gJy4vc3VtJ1xuXG5jb25zdCBpQW1VbnVzZWRBbmRSZW1vdmVkID0gJ2lBbVVudXNlZEFuZFJlbW92ZWQnXG5cbmZ1bmN0aW9uIHNvbWVmdW5jKCkge1xuICBcbn1cblxuY29uc3QgaUFtVW51c2VkID0gc29tZWZ1bmMoKVxuXG5cbmNvbnNvbGUubG9nKCdoZWxsbycpIn1dLFsiZXNidWlsZC5jb25maWcuanNvbiIseyJjb250ZW50Ijoie1xuICBcImZvcm1hdFwiOiBcImNqc1wiLFxuICBcImNkblVybFwiOiBcImh0dHBzOi8vY2RuLnNreXBhY2suZGV2XCJcbn0ifV0sWyJzdW0udHMiLHsiY29udGVudCI6ImV4cG9ydCBjb25zdCBzdW0gPSAoYTogbnVtYmVyLCBiOiBudW1iZXIpID0+IGEgKyBiXG5jb25zb2xlLmxvZygnaGVsbG8nKSJ9XSxbImZpbGUtMy50cyIseyJjb250ZW50IjoiIn1dXQ==)

poteboy:
@naruaway 
You're absolutely right, it's much easier to read code without unused variables. Now I'm considering whether setting it to 'warning' rather than 'off' would be a good compromise. I'm keen to hear what @kotarella1110  and @taishinaritomi  think about this

naruaway:
Ok, now I rebased on the latest `main` and disabled `@typescript-eslint/no-unused-vars` :thumbs_up:

Also I included the following changes this time:

- Adding `eslint:recommended` only revelaed 7 additional issues so this should not add much noise. I included this as well since this includes all the officially recommended rules which is unrelated to TypeScript
- I changed the lint commands to the following: `eslint './packages/**/*.{js,ts,jsx,tsx}' --report-unused-disable-directives --max-warnings 0`, which includes `report-unused-disable-directives` and `--max-warnings 0`. The former is useful to automatically remove unused "eslint-disable" comments (see the actual example in this PR) and the latter is useful to let CI fail for ESLint "warnings". If we simply ignore "warnings" there is no point to use warnings :)
- I added `ignorePatterns` in `.eslintrc.js` to avoid accidentally running ESLint for dist directories

naruaway:
@poteboy, personally I would not recommend to rely on "warning" since everyone will just ignore "warning" eventually and huge amount of warnings will be just unnecessary noise **if CI just passes with warnings**.

As I just posted [here](https://github.com/poteboy/kuma-ui/pull/238#issuecomment-1646405503) I actually just added `--max-warnings 0` so that even single warning should fail the CI 😅  Note that this does not mean we should not use "warning" in config. We can utilize "warning" in different way. For example, configuring IDE's real time linting to only ignore "warning" can help quick prototyping.

Note that It's also possible to merge this PR first and later we can remove "@typescript-eslint/no-unused-vars: off" from the PR at some point. My primary motivation for this PR is to introduce type related rules such as [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md), which can reveal real serious bugs

packages/babel-plugin/package.json
```diff
@@ -32,8 +32,8 @@
     "build": "tsup --config ../../tsup.config.ts",
     "typecheck": "tsc --noEmit --composite false",
     "test": "vitest run",
-    "lint": "eslint './src/**/*.{js,ts,jsx,tsx}'",
-    "lint:fix": "eslint --fix './src/**/*.{js,ts,jsx,tsx}'"
+    "lint": "eslint './src/**/*.{js,ts,jsx,tsx}' --report-unused-disable-directives --max-warnings 0",
```
naruaway, line 35: (note: I am interested in how we can share the same command across packages but not sure about clean way; shell script? 😅 )

naruaway:
Anyway probably it's a good idea to have `"@typescript-eslint/no-unused-vars": "off",` in this PR just to reduce the amount of diff 👍

naruaway:
Anyway probably it's a good idea to have `"@typescript-eslint/no-unused-vars": "off",` in this PR just to reduce the amount of diff 👍

packages/compiler/src/extractor/extract.ts
```diff
@@ -77,6 +82,7 @@ export const extractProps = (
   // However, we do not want to apply the Box theme in those cases.
   if (componentName === "Box" && isDefault) {
     for (const prop in componentVariantProps) {
+      // eslint-disable-next-line no-prototype-builtins -- FIXME
       if (componentVariantProps.hasOwnProperty(prop)) {
```
naruaway, line 86: Note: this shows`eslint:recommended` (specifically [this rule](https://eslint.org/docs/latest/rules/no-prototype-builtins)) can be useful to have 🎉  (well this is just an unlikely edge case but  `componentVariantProps` might. be `{hasOwnProperty: "something"}`, still better to avoid this pattern)

If we can only support above Node.js v16 recent minor, we can just use [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn)

If we care old Node.js we can use `Object.prototype.hasOwnProperty.call(componentVariantProps, prop);`


(I can create follow up PR to fix some of FIXME comments added in this PR. It's better to make sure that this PR does not change any runtime behavior)

naruaway:
Note: I'll rebase and resolve conflict when this PR gets actively reviewed by the maintainers since the conflict will most likely happen when we touch some files

poteboy:
@naruaway 
If I remember correctly, you're working on a follow-up PR for this, right? If so, I guess it's ready to merge once all the conflicts are resolved. What do you guys think? @kotarella1110 @taishinaritomi

packages/babel-plugin/package.json
kotarella1110: I don't know of such a method either...
There are cases where different commands need to be set for each package, so I think it's fine as it is.

kotarella1110:
I have reviewed all the PRs. It looks fine, so please resolve conflicts and prepare for the merge :pray:

taishinaritomi:
I have reviewed the PR. It seems to be fine.

kotarella1110 (approved):
LGTM 🎉

packages/compiler/src/extractor/extract.ts
naruaway: Follow up MR https://github.com/poteboy/kuma-ui/pull/244

PR #240, opened 2023-07-22
Enhance Performance by Optimizing Compiler for Static HTML Conversion

poteboy:
This PR introduces an enhancement to the compiler to improve performance, particularly when using static styles. 

The performance bottleneck was identified in a [css-in-js-perfs](https://github.com/naruaway/css-in-js-perfs), which indicated that Kuma is slower than the baseline, even when all styles can be statically extracted.

This issue arises because Kuma doesn't fully leverage the benefits of static styles. Specifically, components such as `<Flex as='section' fontSize={24} />` don't get fully converted to plain HTML elements during compilation, even though they should in theory be able to, given that all styles are static.

To address this, the Kuma compiler is updated to transform components into their HTML equivalents directly in the AST when possible. For instance, a `<Flex>` component would be converted to a `<section>`, based on the `as` prop, or to the default HTML element for that component. 

However, this transformation can only be applied when no runtime-affecting prefix utility props are present, since these would necessitate runtime processing and therefore an intermediary component. Additionally, care has been taken to handle edge cases such as spread syntax.

With these updates, we anticipate a significant improvement in performance, bringing it closer to the baseline and providing a more efficient user experience.


packages/compiler/src/optimizer/optimize.ts
```diff
@@ -0,0 +1,68 @@
...
+    return true;
+  });
+
+  if (!isOptimizable) return;
+
+  const rawHTMLTag = (() => {
+    const tag = defaultComponentTag[componentName];
+    if (!!as && typeof as === "string") {
+      jsxElement.getAttribute("as")?.remove();
+      return (as as string).replace(/['"`]/g, "");
```
naruaway, line 33: I think it's better to make sure `as` does not include things like `'` at this point rather than adding special handling here, which is error prone

naruaway:
`src/optimizer/optimize.ts` looks good to me at high level :tada: but it is not yet handling `<Flex as='section'>` pattern mentioned in the PR description.
Note that in theory we can even handle custom component like the following (not sure whether we want to handle these in this PR though):

- `<Box as={MyCustomComponent} name={x} />` -> `<MyCustomComponent name={x} >`
- `<Box as={someExpression} name={x} />` -> `const C = someExpression; <C name={x} >`

poteboy:
@taishinaritomi @kotarella1110 cc @naruaway 
It's ready for review. Meanwhile, I will measure the benchmark to compare how much this change has improved the speed of the Kuma component.

packages/compiler/src/optimizer/optimize.ts
```diff
@@ -0,0 +1,115 @@
...
+  try {
+    if (Node.isJsxOpeningElement(jsxElement)) {
+      const jsxElementParent = jsxElement.getParentIfKind(
+        SyntaxKind.JsxElement
+      );
+      if (jsxElementParent) {
+        jsxElementParent
+          .getOpeningElement()
+          .getTagNameNode()
+          .replaceWithText(newTagName);
```
poteboy, line 86: This is fine:
```tsx
<Box as='section' >
   <Box as='div' />
</Box>
```

This also works fine:
```tsx
<Box>
   <Box as='div' />
</Box>
```

And this is fine too:
```tsx
<Box>
   <Box/>
</Box>
```

However, it somehow fails only when the parent component has an 'as' prop and the child component doesn't:
```tsx
<Box as='section'>
   <Box/>
</Box>
```

poteboy:
## Performance Optimization Report
### Rendering Duration (actualDuration):
**Before Optimization**: 350.4ms
**After Optimization**: 195.1ms
**Improvement**: Reduced by 44.3%

### Base Rendering Duration (baseDuration):
**Before Optimization**: 276.2ms
**After Optimization**: 122.7ms
**Improvement**: Reduced by 55.6%

### Commit Duration (commitTime):
**Before Optimization**: 741.9ms
**After Optimization**: 579.7ms
**Improvement**: Reduced by 21.9%

Overall Assessment:
Based on the metrics from rendering 1,000 instances of the `Hero` component, we've seen a significant speed increase post-optimization. The base rendering time, in particular, has improved by an impressive **55.6%**. This optimization is expected to greatly enhance the user experience.

packages/compiler/src/optimizer/optimize.ts
yukukotani: The last pattern works in my environment 🤔 

output:
```tsx
<Box as="section" className={"kuma-1218924623"}>
  <div className={"kuma-1218924623"} />
</Box>
```

packages/compiler/src/optimizer/optimize.ts
poteboy: @naruaway 
Still a bit ad-hoc but fixed it [6375b38](https://github.com/kuma-ui/kuma-ui/pull/240/commits/6375b38f84d375d118604f52d342a1550fd2033b)

poteboy:
@naruaway 
> Note that in theory we can even handle custom component like the following (not sure whether we want to handle these in this PR though)

This is definitely worth trying. However, it requires a lot of modifications to other code, so I will submit it in another PR. 🙏

poteboy:
After addressing the areas pointed out by @yukukotani , the rendering time was reduced to a quarter of its original duration xD

Here are the results:

## Rendering Duration (actualDuration):

- Before Optimization: 350.4ms
- After Optimization (based on @yukukotani's feedback): 140.8ms
- Improvement: Reduced by 59.8%

## Base Rendering Duration (baseDuration):
- Before Optimization: 276.2ms
- After Optimization (based on @yukukotani's feedback): 72.6ms
- Improvement: Reduced by 73.7%

##  Commit Duration (commitTime):
- Before Optimization: 741.9ms
- After Optimization (based on @yukukotani's feedback): 204.5ms
- Improvement: Reduced by 72.4%

taishinaritomi (approved):

PR #243, opened 2023-07-26
docs(website): improve hybrid example for static branching

kotarella1110:
Thanks to the amazing PR by @naruaway at https://github.com/poteboy/kuma-ui/pull/233, static styles will now be generated for conditional expression props. As a result, we needed to make changes to the Example using the hybrid approach that we have featured on the website.

I plan to merge this after that PR has been merged.


website/src/pages/docs/Recepies/Dynamic.mdx
```diff
@@ -13,12 +13,12 @@ For example, in the following code snippet, `p={12}` can be statically interpret
 import React from "react";
 import { Box, Button } from "@kuma-ui/core";
 
-export function Client() {
-  const [checked, toggle] = React.useReducer((state) => !state, false);
+function App() {
+  const [color, setColor] = React.useState("blue");
```
naruaway, line 17: I have no concrete idea in my mind yet but this example might be "bad example" since if we just use conditional expression instead here, we can express exact the same behavior (I mean, the example before this PR) in cleaner code and that will be converted to static CSS, which is better in every aspect.

People tend to use examples for copy-pasting patterns so if we can come up with more  practical example here rather than "toggle with unnecessary indirection", it would be nice!

naruaway: https://github.com/callstack/linaria/blob/master/docs/DYNAMIC_STYLES.md#inline-styles can be a good source of inspiration

> However, it's not possible to use inline styles with pseudo-selectors or media queries.

So something like the following can be a better example?
```tsx
function Item({ index, children }) {
  return (
    <Box transform={[`translateX(${index * 100}%)`, `translateX(${index * 80}%)`]}>
      {children}
    </Box>
  );
}
```

This is not possible with `style` props :) So there is a reason to use libraries with dynamic capability such as Kuma UI for this case

PR #244, opened 2023-07-27
Eslint followups

naruaway:
This PR is small follow ups after https://github.com/poteboy/kuma-ui/pull/238. Each commit message should describe what it does.

I just realized our GitHub actions do not run `pnpm run lint` so new PRs with failing lint are being merged 😅  If the maintainers are fine, I can try adding lint to the Github Actions in this PR as well 

## The usage of Object.hasOwn
According to [Node.js releases](https://nodejs.dev/en/about/releases/), Node.js v16 is the "Maintenance LTS" and the current active LTS is v18.
Also there should be no reason not to update Node.js minor versions.
So I think it's reasonable to assume that people are using Node.js newer than 16.9.0, which means we can use new APIs such as [Object.hasOwn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) for plugins (runtime is different, which runs in browsers!) I think it's better to clarify supported Node.js versions in docs and engines field in package.json but it's out of scope for this PR

Or actually we can just drop hasOwn checks if we believe in modern environment, no one pollutes Object.prototype :)

### DIFF
--- a/packages/babel-plugin/src/components/processComponents.ts
+++ b/packages/babel-plugin/src/components/processComponents.ts
@@ -19,7 +19,9 @@ export const processComponents = (
       const openingElement = path.node.openingElement;
       if (t.isJSXIdentifier(openingElement.name)) {
         const name = openingElement.name.name;
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- FIXME
         const componentType =
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- FIXME
           componentList[
             importedStyleFunctions[name] as keyof typeof componentList
           ];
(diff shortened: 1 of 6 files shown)


poteboy:
> If the maintainers are fine, I can try adding lint to the Github Actions in this PR as well

This would be very helpful

naruaway:
@poteboy, cool, then I just added "lint" GitHub action in this PR and confirmed that when lint is failing it blocks the PR by testing "intentionally broken commit" in this PR (already removed now):

fail-in-pr
fail

poteboy (approved):
LGTM!

Issue #150, opened 2023-06-26
unsupported CSS prop list


masa5555:
### -moz-\_

- [ ] ~~-moz-float-edgeNon-standardDeprecated~~
- [ ] ~~-moz-force-broken-image-iconNon-standardDeprecated~~
- [ ] ~~-moz-image-regionNon-standard~~
- [ ] ~~-moz-orientNon-standard~~
- [ ] ~~-moz-user-focusNon-standard~~
- [ ] ~~-moz-user-inputNon-standardDeprecated~~

### -webkit-\_

- [ ] ~~-webkit-box-reflectNon-standard~~
- [ ] -webkit-line-clamp
- [ ] ~~-webkit-mask-attachmentNon-standard~~
- [ ] ~~-webkit-mask-box-imageNon-standard~~
- [ ] ~~-webkit-mask-compositeNon-standard~~
- [ ] ~~-webkit-mask-position-xNon-standard~~
- [ ] ~~-webkit-mask-position-yNon-standard~~
- [ ] ~~-webkit-mask-repeat-xNon-standard~~
- [ ] ~~-webkit-mask-repeat-yNon-standard~~
- [ ] ~~-webkit-overflow-scrollingNon-standard~~
- [ ] ~~-webkit-tap-highlight-colorNon-standard~~
- [ ] -webkit-text-fill-color
- [ ] ~~-webkit-text-securityNon-standard~~
- [ ] -webkit-text-stroke
- [ ] -webkit-text-stroke-color
- [ ] -webkit-text-stroke-width
- [ ] ~~-webkit-touch-calloutNon-standard~~

###

- [x] accent-color

### align-\_

- [x] align-content
- [x] align-items
- [ ] align-self
- [ ] align-tracksExperimental

###

- [ ] all

### animation-\_

- [ ] animation
- [ ] animation-composition
- [ ] animation-delay
- [ ] animation-direction
- [ ] animation-duration
- [ ] animation-fill-mode
- [ ] animation-iteration-count
- [ ] animation-name
- [ ] animation-play-state
- [ ] animation-timelineExperimental
- [ ] animation-timing-function

###

- [ ] appearance
- [ ] aspect-ratio
- [ ] backdrop-filter
- [ ] backface-visibility

### background-\_

- [x] background
- [ ] background-attachment
- [ ] background-blend-mode
- [ ] background-clip
- [x] background-color
- [ ] background-image
- [ ] background-origin
- [ ] background-position
- [ ] background-position-x
[...]

PR #245, opened 2023-07-30
add: properties related to border

nojiritakeshi:
The following styles were added.
```
  borderX: "border-left,border-right",
  borderY: "border-top,border-bottom",
  borderTopLeftRadius: "border-top-left-radius",
  borderTopRightRadius: "border-top-right-radius",
  borderBottomLeftRadius: "border-bottom-left-radius",
  borderBottomRightRadius: "border-bottom-right-radius",
  borderTopStyle: "border-top-style",
  borderBottomStyle: "border-bottom-style",
  borderLeftStyle: "border-left-style",
  borderRightStyle: "border-right-style",
  borderTopWidth: "border-top-width",
  borderBottomWidth: "border-bottom-width",
  borderLeftWidth: "border-left-width",
  borderRightWidth: "border-right-width",
  borderStart: "border-inline-start",
  borderEnd: "border-inline-end",
  borderStartWidth: "border-inline-start-width",
  borderEndWidth: "border-inline-end-width",
  borderStartStyle: "border-inline-start-style",
  borderEndStyle: "border-inline-end-style",
  borderStartRadius: "border-top-left-radius,border-bottom-left-radius",
  borderEndRadius: "border-top-right-radius,border-bottom-right-radius",
```


packages/system/src/border.ts
```diff
@@ -1,12 +1,87 @@
...
+     */
+    borderEndWidth: CSSValue<"borderInlineEndWidth", true>;
+    /**
+     * @see borderInlineStartStyle
+     */
+    borderStartStyle: CSSValue<"borderInlineStartStyle">;
+    /**
+     * @see borderInlineEndStyle
+     */
+    borderEndStyle: CSSValue<"borderInlineEndStyle">;
```
poteboy, line 41: I don't believe that `borderEndStyle` should accept a numerical value. This concern also extends to some of the other styles added in this PR.

nojiritakeshi: @poteboy 
I don't think borderEndStyle accepts numeric values

poteboy: Oh, what I actually meant by that is we should filter the third argument of `applyResponsiveStyles` based on whether or not the property accepts a numerical value.

nojiritakeshi: I fixed what you pointed out. 
please confirm.

poteboy (approved):

masa5555:
#150

PR #248, opened 2023-07-31
refactor: move tagged template expression from babel-plugin to compiler

kotarella1110:
We have decided to migrate the static extraction logic of TaggedTemplateExpression from babel-plugin to the compiler.
This means that the elements like styled and css API, which were previously statically extracted using babel, will now be extracted using ts-morph. As a result, we will be able to perform static extraction using ts-evaluate in the future, enabling smarter code processing.
This PR can be seen as a preparation to support interpolations and keyframes.

### DIFF
--- a/packages/compiler/src/processTaggedTemplateExpression.ts
+++ b/packages/compiler/src/processTaggedTemplateExpression.ts
@@ -1,0 +1,53 @@
+import {
+  Node,
+  SyntaxKind,
+  TaggedTemplateExpression,
+} from "ts-morph";
+import { sheet } from "@kuma-ui/sheet";
+
+const parseCssTemplate = (cssTemplateLiteral: Node) => {
+  if (Node.isNoSubstitutionTemplateLiteral(cssTemplateLiteral)) {
+    const cssString = cssTemplateLiteral.getLiteralText();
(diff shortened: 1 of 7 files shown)


packages/compiler/src/processTaggedTemplateExpression.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+  // styled("xxx")``
+  else if (
+    Node.isCallExpression(tag) &&
+    tag.getExpressionIfKind(SyntaxKind.Identifier)?.getText() ===
+      bindings["styled"]
+  ) {
+    const cssTemplateLiteral = node.getTemplate();
+    const className = parseCssTemplate(cssTemplateLiteral);
```
poteboy, line 32: This function seems to be used with two APIs, `styled` and `css`, so the naming of `parseCssTemplate` feels a little off.

kotarella1110: renamed to `extractClassName`: https://github.com/kuma-ui/kuma-ui/pull/248/commits/ae0ffa6ae4338ceedf0c50b4f6ff9fdba42f797b

poteboy (approved):
🫰

Issue #258, opened 2023-08-03
Add longhand properties


yukukotani:
Shorthand properties such as `bg` or `m` exists in StyleProps but long-hand ones such as `background` or `margin` doesn't. Is this expected, or missing?

I'd work on this if it's just missing.

poteboy:
Just missing, go for it xD @yukukotani

yukukotani:
#259

PR #259, opened 2023-08-04
Support long-hand props

yukukotani:
#258


yukukotani:
@poteboy Let me align the approach of implementation before working on other properties. I think it's better to list shorthands naively like this PR while we can generalize the shorthand application by like ef5b4e84c31f52f35cb528ddbb329634d40f3a16. It will be too complicated because of multi props shorthands such as `mx`. What do you think?

poteboy:
@yukukotani 
Thanks for the PR! Checked out your generalized version as well. It's smart, no doubt, but since we've got only a few props that use both short and long, I reckon sticking with the straightforward approach in this PR could be easier to handle. 
What do you guys think? @taishinaritomi @kotarella1110

kotarella1110:
I agree that considering the limited number of props using both short and long forms, sticking with the straightforward approach in this PR seems like a better choice 👍
While the generalized approach is intriguing, prioritizing simplicity for now appears to be more manageable.

packages/system/src/keys.ts
yukukotani, line 1: You can see the list of new props here

packages/system/src/background.ts
```diff
@@ -10,57 +10,77 @@ type AddProperty<T, T2> = {
...
+      | "backgroundOrigin"
+    > &
+      CSSProperties<
+        "backgroundPositionX" | "backgroundPositionY" | "backgroundSize",
+        true
+      > & {
+        /**
+         * @see backgroundImage
+         */
+        bgImage: CSSValue<"backgroundImage"> | AutoPrefix;
```
yukukotani, line 28: Those `AutoPrefix` looks needless since it's in `AddProperty` but keeping it left as it's out of scope of this PR.

yukukotani:
Ready for getting review!

packages/system/src/background.ts
poteboy: Oh thanks. Honestly I don't know when and where this came from lol

poteboy (approved):
Looks good to me!

poteboy:
@yukukotani 
Sorry, one small thing. Can you add a changeset to this PR before merging?

yukukotani:
@poteboy Done :+1: 397cedb

PR #265, opened 2023-08-05
Run changeset status on CI

takurinton:
https://github.com/kuma-ui/kuma-ui/issues/262#issuecomment-1666370838

> So, I'm thinking if a PR modifies anything under packages/ and doesn't have a changeset, we'll fail the CI. We'll also update the CONTRIBUTING.md to include instructions for adding a patch update changeset.

- [x] Run changeset status on CI
- [x] Fix `CONTRIBUTING.md`

PR #267, opened 2023-08-06
Fix workflows

takurinton:
https://github.com/kuma-ui/kuma-ui/pull/265#issuecomment-1666664520

- `.github/workflows/main.yml` should only be run when `packages/**` changes
- Exclude the `experimental` branch as it is not used

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ on:
       - main
     paths:
       - "packages/**"
-      - ".github/workflows/main.yml"
+      # - ".github/workflows/main.yml"
   workflow_dispatch:
 jobs:
   test:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,11 +3,9 @@ on:
   push:
     branches:
       - main
-      - experimental
   pull_request:
     branches:
       - main
-      - experimental
     paths:
       - "packages/**"
       # - ".github/workflows/main.yml"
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -3,11 +3,12 @@ on:
...
     branches:
       - main
-      - experimental
   pull_request:
     branches:
       - main
-      - experimental
+    paths:
+      - "packages/**"
+      - ".github/workflows/main.yml"
```
takurinton, line 11: In addition to packages, CI should also run if this file itself changes.

poteboy:
@takurinton 
Thanks for the PR!
One question: if GitHub Actions only run when changes are made to the packages/** directory, what happens when we want to modify the website? It seems like changes to the website should also trigger a build? Correct me if I'm wrong xD

takurinton:
@poteboy 

I'm thinking that the main.yml CI doesn't need to be running since the vercel CI does the website changes.

poteboy:
Oh that's right lol

takurinton:
Also, since this part remains, it will be deployed to the production!!

https://github.com/kuma-ui/kuma-ui/pull/267/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R2-R5

takurinton:
Ready for review

poteboy (approved):
🐜1️⃣0️⃣

PR #269, opened 2023-08-06
Enable pnpm cache on CI

yukukotani:
Now, installing dependencies spent 10-15sec, becoming one-third!

Not cached: https://github.com/kuma-ui/kuma-ui/actions/runs/5774108606/job/15650715632?pr=269
Cached: https://github.com/kuma-ui/kuma-ui/actions/runs/5774120528/job/15650743609

The doc about pnpm cache on GitHub Actions here: https://pnpm.io/continuous-integration#github-actions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,14 +20,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v1
-      - name: Setup node ${{ matrix.node_version }}
-        uses: actions/setup-node@v1
-        with:
-          node-version: ${{ matrix.node_version }}
 
       - uses: pnpm/action-setup@v2
         with:
           version: 8.5.1
+
+      - name: Setup node ${{ matrix.node_version }}
+        uses: actions/setup-node@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -20,15 +20,17 @@ jobs:
...
-        uses: actions/setup-node@v1
-        with:
-          node-version: ${{ matrix.node_version }}
 
       - uses: pnpm/action-setup@v2
         with:
           version: 8.5.1
 
+      - name: Setup node ${{ matrix.node_version }}
+        uses: actions/setup-node@v3
```
yukukotani, line 29: We need a major update of setup-node to use cache option. But I think it's ok since it's already used in release.yml with same options.

yukukotani:
Please let me know if internal changes like this also require changeset.

poteboy:
@yukukotani 
Thank you so much for improving the workflow! As for a changeset, it's not required, but you're welcome to add one 👍

poteboy (approved):

PR #271, opened 2023-08-06
 pnpm install in workflows uses composite action

takurinton:
The current configuration of github actions is running pnpm install on each job.
This can be made common by using composite.

https://docs.github.com/en/actions/creating-actions/creating-a-composite-action


.github/actions/setup-pnpm/action.yml
```diff
@@ -0,0 +1,23 @@
+name: setup-pnpm
+description: "This is a composite GitHub Action that checks out the repo, sets up pnpm, node and installs the project's dependencies."
```
yukukotani, line 2: nits: It seems that this action doesn't include checkout step
```suggestion
description: "This is a composite GitHub Action that sets up pnpm, node and installs the project's dependencies."
```

takurinton: Thanks!!

poteboy (approved):
🫰

PR #272, opened 2023-08-06
feat: support  font theme tokens

taishinaritomi:
- fonts
- fontSizes
- fontWeights
-  lineHeights
- letterSpacings


packages/core/src/themeTokens.ts
```diff
@@ -0,0 +1,22 @@
+import { Theme } from "./theme";
+import { FlattenObject, NestedObject, Pretty } from "./utils/object";
+import { If, IsNever, _String, Stringify } from "./utils/types";
+import { Tokens } from "@kuma-ui/sheet";
+
+type ThemeTokens<T extends string> = Theme extends Record<T, unknown>
+  ? Theme[T]
+  : never;
+
+export type InputThemeTokens = Partial<Record<Tokens, NestedObject<string>>>;
```
kotarella1110, line 10: I thought tokens like `breakpoints` and `fontSize` should also accept `number` in theme.Depending on the property, it may be `string` or `string | number`.
Therefore, I thought it necessary to define the type of each token.

packages/system/src/font.ts
```diff
@@ -61,9 +67,46 @@ export const font = (props: FontProps): ResponsiveStyle => {
...
+        if (property === "font-family") {
+          let newValue = value;
+          if (userTheme.fonts) {
+            for (const key in userTheme.fonts) {
+              if (value === key) {
+                newValue = userTheme.fonts[key];
+                break;
+              }
+            }
+          }
```
kotarella1110, line 82: I think it can be refactored as follows:

```suggestion
          if (value in (userTheme.fonts ?? {})) {
            newValue = userTheme.fonts[value];
          }
```

packages/system/src/font.ts
```diff
@@ -61,9 +67,46 @@ export const font = (props: FontProps): ResponsiveStyle => {
...
+        } else if (property === "font-size") {
+          let newValue = toCssUnit(value);
+          if (userTheme.fontSizes) {
+            for (const key in userTheme.fontSizes) {
+              if (value === key) {
+                newValue = toCssUnit(userTheme.fontSizes[key]);
+                break;
+              }
+            }
+          }
```
kotarella1110, line 93: same as above

```suggestion
          if (value in (userTheme.fontSizes ?? {})) {
            newValue = toCssUnit(userTheme.fontSizes[value]);
          }
```

packages/system/src/font.ts
```diff
@@ -61,9 +67,46 @@ export const font = (props: FontProps): ResponsiveStyle => {
...
+        } else if (property === "font-weight") {
+          let newValue = value;
+          if (userTheme.fontWeights) {
+            for (const key in userTheme.fontWeights) {
+              if (value === key) {
+                newValue = userTheme.fontWeights[key];
+                break;
+              }
+            }
+          }
```
kotarella1110, line 104: same as above

```suggestion
          if (value in (userTheme.fontWeights ?? {})) {
            newValue = userTheme.fontWeights[value];
          }
```

poteboy (approved):
🚀

PR #276, opened 2023-08-07
fix: Replace broken generateKey with fast-json-stable-stringify for pseudo props

### LINKED ISSUE #246: Style cache is broken for pseudo props
Using stricter TypeScript typing and type-aware ESLint rule has revealed that style cache stringify logic has a bug, which is accidentally converting object into `[object Object]` string.

See https://github.com/poteboy/kuma-ui/pull/233#discussion_r1278577782 for more details.

We need to either fix the logic or we should rely on well tested libraries such as [fast-json-stable-stringify](https://www.npmjs.com/package/fast-json-stable-stringify)

naruaway:
Closes #246 

## Problem

In short, Kuma UI's pseudo props support is basically broken.

Before this PR,  `generateKey` was generating something like `_hover:[object Object]|p:2`. After this PR, it uses `fast-json-stable-stringify` and it generates `{"_hover":{"color":"red"},"p":2}`, which includes nested values.
This is critical when using pseudo props. Because of this, when using pseudo props, most of the time styling was broken since actual values of pseudo stlying is not taken into account when taking a hash for style caching.

I used the following example to confirm the bug:
```tsx
import { Input } from "@kuma-ui/core";

function Demo() {
  return (
    <>
      <Input m={2} _valid={{ bgColor: 'red' }} />
      <Input m={2} _valid={{ bgColor: 'blue' }} />
    </>
  );
}
```
### Before this PR
before

Both input elements are red and "bgColor: blue" is completely ignored

### After this PR
after

Colors are expected

## Solution

Use `fast-json-stable-stringify` instead of using our own implementation of generateKey.

Note that this logic can be hot execution path when Kuma UI fails to extract static styles.
For example, emotion is tying to optimize [the whole "serialize styles" thing](https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/serialize/src/index.js#L310-L388) and in my experience of using Emotion in large scale website, this was a heavy bottleneck. We might want to revisit this choice in the future but we should fix a bug first.

## The choice of fast-json-stable-stringify

I ran [this benchmark](https://github.com/epoberezkin/fast-json-stable-stringify#benchmark) in my MacBook:
```
fast-json-stable-stringify x 61,432 ops/sec ±0.20% (101 runs sampled)
json-stable-stringify x 48,248 ops/sec ±0.12% (97 runs sampled)
fast-stable-stringify x 65,665 ops/sec ±0.10% (97 runs sampled)
faster-stable-stringify x 55,941 ops/sec ±1.77% (96 runs sampled)
The fastest is fast-stable-stringify
```

For now I would choose "fast-json-stable-stringify" for since I found it's used in [`@jest/transform`](https://github.com/facebook/jest/blob/0fd5b1c37555f485c56a6ad2d6b010a72204f9f6/packages/jest-transform/src/ScriptTransformer.ts#L14-L14).
Also note that this relies on the fact **Kuma UI never relies on prop ordering even for nested cases**.

### DIFF
--- a/packages/core/src/components/Box/react/utils.ts
+++ b/packages/core/src/components/Box/react/utils.ts
@@ -1,5 +1,6 @@
 import { isStyledProp, isPseudoProps, StyleGenerator } from "@kuma-ui/system";
 import { BoxProps } from "./types";
+import { stableStringify } from "../../../utils/stableStringify";
 
 function isDynamicProp(key: string) {
   if (isStyledProp(key) || isPseudoProps(key) || key === "variant") {
@@ -41,23 +42,9 @@ const styleCache: {
   [key: string]: { className: string; css: string } | undefined;
 } = {};
 
-/**
- * Generates a unique key for props, aiding cache efficiency.
- * Incurs O(n log n) cost due to sorting, but it's acceptable given the
- * expensive nature of StyleGenerator's internals.
- */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any -- FIXME
-function generateKey(props: Record<string, any>) {
-  return Object.entries(props)
-    .filter(([, value]) => value !== undefined)
-    .sort((a, b) => a[0].localeCompare(b[0]))
-    .map(([key, value]) => `${key}:${value}`)
-    .join("|");
-}
-
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- FIXME
 export function getCachedStyle(dynamicProps: Record<string, any>) {
-  const key = generateKey(dynamicProps);
+  const key = stableStringify(dynamicProps);
   let generatedStyle = styleCache[key];
   // If the result isn't in the cache, generate it and save it to the cache
   if (!generatedStyle) {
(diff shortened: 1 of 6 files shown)


naruaway:
Closes #246

packages/core/src/components/Box/react/utils.ts
```diff
@@ -41,23 +42,9 @@ const styleCache: {
...
-    .filter(([, value]) => value !== undefined)
-    .sort((a, b) => a[0].localeCompare(b[0]))
-    .map(([key, value]) => `${key}:${value}`)
-    .join("|");
-}
-
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- FIXME
 export function getCachedStyle(dynamicProps: Record<string, any>) {
-  const key = generateKey(dynamicProps);
+  const key = stableStringify(dynamicProps);
```
naruaway, line 47: Note: I am not sure whether it is intentional to have style cache logic in this file as well

naruaway:
Discussed with the core maintainers, we decided to vendor fast-json-stable-stringify in the codebase to avoid adding new package dependency.
Since we need it for two different packages, for now we duplicate the implementation with code comment. In the future we should come up with a way to share intenal utils across packages without exposing them to end users

PR #277, opened 2023-08-07
ci: Run build before lint to make type-aware rules work properly

naruaway:
I just realized that running `pnpm lint` without build artifacts behaves in unexpected way since internal packages types are not discoverable. To avoid running prettier check for dist/, I also added .prettierignore

~Note that I think most of projects are relying on ESLint to make sure all the files are prettiered but this repo does not seem to integrate Prettier with ESLint in that way, here I am assuming that is intentional~ **UPDATE**: It was [pointed out](https://github.com/kuma-ui/kuma-ui/pull/277#discussion_r1285490305) that we should not configure ESLint to complain about Prettier


poteboy:
resolves #246

.github/workflows/main.yml
```diff
@@ -43,12 +43,12 @@ jobs:
         with:
           node_version: 16
 
+      - name: Build packages
+        run: pnpm build
+
       - name: Run lint
         run: pnpm lint
 
-      - name: Run format
```
poteboy, line 49: I believe this code was added by @yukukotani  yesterday. Could you confirm if it's appropriate to delete it?

yukukotani: Sorry, I don't get why this should be removed.

> Note that I think most of projects are relying on ESLint to make sure all the files are prettiered but this repo does not seem to integrate Prettier with ESLint in that way, here I am assuming that is intentional

IMO, as @naruaway said, our ESLint configuration doesn't integrate with Prettier, so we have to manually check the format using `pnpm format`.

yukukotani: In addition, integrating Prettier with ESLint is generally not recommended.

ref: https://prettier.io/docs/en/integrating-with-linters#notes

poteboy: Based on what @yukukotani said, I'm inclined to think that this code should remain as it is. If there's any particular reason for removing it, could you please share it with me? @naruaway

naruaway: Thanks for the info, this was my mistake. I initially thought ESLint is covering Prettier so I removed this line but then I realized our ESLint is not configured in that way as I mentioned in PR description. And then I forgot to push the change to restore this "Run format" 😭 

I now restored it and rebased 👍 

> integrating Prettier with ESLint is generally not recommended.

Oh good to know! And the rationale explained in the Prettier doc totally makes sense

poteboy (approved):

Issue #256, opened 2023-08-03
Implementation of ForwardRef needed for proper handling of refs


poteboy:
We have received reports from several users indicating that Kuma UI is not fully compatible with libraries like React Hook Form. Upon investigation, it was found that the incompatibility arises from the lack of `ForwardRef` implementation in our components.

`ForwardRef` in React allows a component to access a `ref `from its parent, which is essential for functionalities like focusing an input field, measuring the dimensions of an element, etc. Libraries like React Hook Form rely heavily on this feature to associate form data with specific form elements.

By implementing `ForwardRef` in our components, we can ensure better interoperability with such libraries and provide a more robust development experience.

PR #278, opened 2023-08-08
test(core): setup @testing-library/jest-dom and add component test for Box

MH4GF:
## Description

This PR will set up @testing-library/jest-dom to be able to use toBeInTheDocument(), etc. in component testing.

## Motivation

I was going to try this issue https://github.com/kuma-ui/kuma-ui/issues/256, but first I wanted to be able to easily assert that existing usage was not broken.

## How was it tested

As a sample test, I have added the test code for the Box component.
If this plan is acceptable, I will expand the test code for other components 👍🏻

### DIFF
--- a/packages/core/src/tests/vitest.setup.ts
+++ b/packages/core/src/tests/vitest.setup.ts
@@ -1,0 +1,11 @@
+import type { TestingLibraryMatchers } from "@testing-library/jest-dom/matchers";
+import matchers from "@testing-library/jest-dom/matchers";
+import { expect } from "vitest";
+
+declare module "vitest" {
+  interface Assertion<T>
+    extends jest.Matchers<void, T>,
+      TestingLibraryMatchers<T, void> {}
+}
+
+expect.extend(matchers);
--- a/.changeset/kind-socks-boil.md
+++ b/.changeset/kind-socks-boil.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/core": patch
+---
+
+maintenance: setup testing-library/jest-dom and add component test for Box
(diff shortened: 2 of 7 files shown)


packages/core/src/tests/vitest.setup.ts
```diff
@@ -0,0 +1,11 @@
+import type { TestingLibraryMatchers } from "@testing-library/jest-dom/matchers";
+import matchers from "@testing-library/jest-dom/matchers";
+import { expect } from "vitest";
+
+declare module "vitest" {
+  interface Assertion<T>
+    extends jest.Matchers<void, T>,
+      TestingLibraryMatchers<T, void> {}
+}
```
MH4GF, line 9: ref: https://github.com/testing-library/jest-dom/issues/439#issuecomment-1536524120

MH4GF:
@kotarella1110 Oh, I added it! https://github.com/kuma-ui/kuma-ui/pull/278/commits/664ad097a4f4d2fe99cf733c136db3d31e4a98fa

poteboy:
@MH4GF 
Thanks for the PR! Could you modify the tests in `spacer.test.tsx` to use `toHaveStyle` instead of calculating the style from `getComputedStyle`, especially now that we have an extended matcher from `testing-library/jest-dom`?

MH4GF:
@poteboy Of course! But since the scope of this PR is larger, can I use a different PR?
In addition to modifying spacer.test.tsx, I also wanted to move the `cleanup()` to vitest.setup.ts:

https://github.com/kuma-ui/kuma-ui/blob/9a565d75bf22b591ed50c716b6adfdc83f7405a9/packages/core/src/components/Spacer/spacer.test.tsx#L8-L10

poteboy (approved):

PR #283, opened 2023-08-13
Fix the way to calculate processedProps

mehm8128:
I found that when I use both `textAlign='right'`and `marginRight={4}`, the former is not applied.
For example:
The below code,  `aaa` and `bbb` is right-aligned but `ccc` is not aligned.
```
   <div
        style={{
          textAlign: "right",
          marginRight: 4,
        }}
      >
        aaa
      </div>
      <Box textAlign="right">bbb</Box>
      <Box textAlign="right" marginRight={4}>
        ccc
      </Box>
```

So, I tried to fix it, but an error `TypeError: compose is not a function` occurred.
Could you check my code? (both whether my fix is proper fix and why this error occurs)

### DIFF
--- a/.changeset/wise-teachers-build.md
+++ b/.changeset/wise-teachers-build.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/system": patch
+---
+
+Fix the way to calculate processedProps
--- a/packages/system/src/animation.ts
+++ b/packages/system/src/animation.ts
@@ -18,7 +18,7 @@ export type AnimationProps = Partial<
     CSSProperties<"animationIterationCount", true>
 >;
 
-const animationMappings: Record<AnimationKeys, string> = {
+export const animationMappings: Record<AnimationKeys, string> = {
   animation: "animation",
   animationComposition: "animation-composition",
   animationDelay: "animation-delay",
--- a/packages/system/src/background.ts
+++ b/packages/system/src/background.ts
@@ -59,7 +59,7 @@ export type BackgroundProps<AutoPrefix extends string = string & {}> = Partial<
   >
 >;
 
-const backgroundMappings: Record<BackgroundKeys, string> = {
+export const backgroundMappings: Record<BackgroundKeys, string> = {
   backgroundImage: "background-image",
   bgImage: "background-image",
   backgroundPosition: "background-position",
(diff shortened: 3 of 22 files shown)


poteboy:
Thanks for bringing this up. I was able to reproduce the issue on my local environment as well.

I noticed that this issue doesn't occur with `m` or `margin` but specifically with `mr` and `marginRight`. It does seem like something related to `compose.ts`, but it's been 4 months since I last touched that code, so I need to refresh my memory on it 😅

I'll probably take a closer look within the next week.

mehm8128:
Thank you for response.
Yes, this bug occurs when `right` of `margin-right` and `right` of `text-align: right` is overlapping, so when I designate `margin-right` and `justify-content: right`, same issue occurs.
I'm looking forward to your review😃

mehm8128:
Hi, I investigated about this build error(`compose is not a function`).
I looked at built file which is emitted into `packages/system/dist/chunk-*`, and I found that `compose` function is used to define `all` function before its definition in the chunk file(and its definition is arrow function, so hoisting is not performed).
I don't know how to solve it, but I changed the file name from `background.ts` to `xxx.ts`, the build error went away and build succedded.

kotarella1110:
@mehm8128 

> So, I tried to fix it, but an error TypeError: compose is not a function occurred.

The error is caused by a circular reference.
Here is the code for the cause of the circular reference.

https://github.com/kuma-ui/kuma-ui/blob/341430056cff0e4097e0b066726e5c110219eac9/packages/system/src/background.ts#L4

The following fix resolves this error.

```ts
import { toCssUnit } from "./toCSS";
```

> both whether my fix is proper fix and why this error occurs

Your fix is appropriate 👍

poteboy:
Thank you @kotarella1110 🙇‍♂️

mehm8128:
@kotarella1110 Thank you for your review!
I understand why the error was caused and I fixed it.
Could you review my code again?(CC: @poteboy )

kotarella1110 (approved):
Thank you for your contribution! LGTM 🎉

kotarella1110:
@mehm8128  please fix lint error 🙏

poteboy (approved):
Nice work 🙌

PR #284, opened 2023-08-14
Add support for BinaryExpressions in handleJsxExpression

poteboy:
This PR enhances `handleJsxExpression` to support static analysis of BinaryExpressions within JSX props, allowing for arithmetic operations to be evaluated at compile-time.

**Example:**
Previously, expressions like `<Text fontSize={2 + 5} />` weren't statically extractable. With this update, the above JSX will be recognized and evaluated as having a fontSize of `7`.

### DIFF
--- a/packages/compiler/src/collector/expression.ts
+++ b/packages/compiler/src/collector/expression.ts
@@ -1,7 +1,9 @@
-import { Node, ts } from "ts-morph";
+import { Node, SyntaxKind, ts } from "ts-morph";
 import { match } from "ts-pattern";
 
-export const handleJsxExpression = (node: Node<ts.Node>) => {
+export const handleJsxExpression = (
+  node: Node<ts.Node>
+): string | number | boolean | (string | number | undefined)[] | undefined => {
   return (
     match(node)
       // fontSize={24}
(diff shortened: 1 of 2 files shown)


packages/compiler/src/collector/expression.ts
```diff
@@ -37,6 +39,31 @@ export const handleJsxExpression = (node: Node<ts.Node>) => {
...
+      // fontSize={2 + 5}
+      .when(Node.isBinaryExpression, (exp) => {
+        const leftOperand = handleJsxExpression(exp.getLeft());
+        const rightOperand = handleJsxExpression(exp.getRight());
+        const operator = exp.getOperatorToken().getKind();
+        if (
+          typeof leftOperand === "number" &&
+          typeof rightOperand === "number"
+        ) {
+          // FIXME: opted for native switch due to a type conflict between ts-pattern and SyntaxKind which remains unexplained.
```
kotarella1110, line 51: This is a TS limitation.
It would be difficult to resolve type errors with ts-pattern.

https://github.com/gvergnaud/ts-pattern/issues/183

kotarella1110 (approved):
LGTM 🎉

packages/compiler/src/collector/expression.ts
poteboy: I wasn't aware of that! Thank you for letting me know 👍

kotarella1110 (approved):

PR #286, opened 2023-08-15
chore: modified existing test cases with the introduction of `testing-library/jest-dom`

MH4GF:
Minor modifications were made to the test cases in accordance with https://github.com/kuma-ui/kuma-ui/pull/278 .

- test(core): Run cleanup() on all test cases beforeEach
- fix(core): Replace `getComputedStyle` with `toHaveStyle` from `testing-library/jest-dom`


poteboy (approved):
👍

PR #287, opened 2023-08-15
test(core): add some component tests

MH4GF:
Added simple tests for some components of the core package.
I'm adding tests with a policy of fulfilling coverage to some extent.

スクリーンショット 2023-08-16 8 45 50

If this policy is acceptable, it will be reflected in all other components in the next PR.

### DIFF
--- a/.changeset/friendly-islands-eat.md
+++ b/.changeset/friendly-islands-eat.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/core": patch
+---
+
+test(core): add some component tests
--- a/packages/core/src/components/Box/react/box.test.tsx
+++ b/packages/core/src/components/Box/react/box.test.tsx
@@ -4,7 +4,18 @@ import { Box } from ".";
 import React from "react";
 
 describe("Box component", () => {
-  it("should render correctly with given static props", () => {
+  it("should render correctly with no props", () => {
+    // Arrange
+    const props = {};
+
+    // Act
+    render(<Box {...props}>This is a box</Box>);
+
+    // Assert
+    expect(screen.getByText("This is a box")).toBeInTheDocument();
+  });
+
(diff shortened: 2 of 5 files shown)


poteboy:
@MH4GF 
Thank you for your PR! I've taken a look at the PR and have a few points of feedback:
1. The test title mentions checking the "allowed shorthand style props", but it currently only verifies that the `Grid` component renders. It would be beneficial to ensure that each prop is correctly applied and reflects the desired styles.
2. Some of the prop values, like `auto-flow`, might not be appropriate for properties such as `grid`, `gridArea`, etc. Please double-check the valid values for these properties.

MH4GF:
Thank you for your response 😄 

> The test title mentions checking the "allowed shorthand style props", but it currently only verifies that the Grid component renders. It would be beneficial to ensure that each prop is correctly applied and reflects the desired styles.

> By default, the Grid component should have the style display: grid;. It might be good to have a test that verifies this even without any style props provided.

I thought the point you made was valid when I thought of "testing in the kuma ui component.". However, this component of `@kuma-ui/core` does not convert shorthand props, so it seems to be ignored even if props are passed as this component.

With that in mind, it seemed like it would be of little value as a test to perform component testing here.
Rendering after transformation by babel, more like an e2e test, might be more desirable.

What do you think? For my part, I think it might be best to close this PR once it is done.

poteboy:
@MH4GF 
I was a bit puzzled by your comment. When you mentioned `@kuma-ui/core does not convert shorthand props`, could you clarify that? I was under the impression it did.

MH4GF:
@poteboy 
Sorry for the late reply! I'm in the process of reading and debugging the code and trying to understand it better. Please wait a little longer as we will contact you within 3 days.

MH4GF:
Sorry, I will not do this one now. I am rooting of this project, Thanks!

Issue #257, opened 2023-08-03
Support more theme tokens


yukukotani:
[The doc](https://www.kuma-ui.com/docs/Theme/CustomizingTheme#theme-tokens) says that kuma-ui want to support more theme tokens. To follow [Chakra UI](https://chakra-ui.com/docs/styled-system/theme), we need to implement the primitive tokens below:

- [x] colors
- [x] fonts
- [x] fontSizes
- [x] fontWeights
- [x] lineHeights
- [x] letterSpacings
- [x] breakpoints
- [x] space
- [x] sizes
- [x] radii
- [x] zIndices

Panda CSS supports more detail tokens like [borders](https://panda-css.com/docs/theming/tokens#borders).
Let me clarify which tokens should be supported in kuma-ui on this issue, so that I can work on this.

poteboy:
@yukukotani 
In our meeting a week ago, we decided to support `font`, `fontSize`, `fontWeights`, `lineHeights`, and `letterSpacings`. This task has already been assigned to another maintainer, @taishinaritomi , so there might be some overlap with what he's currently working on. It would be a good idea to discuss and split the work with him!
Also, thank you for showing initiative to tackle this issue :)

yukukotani:
@poteboy Thanks for the quick response!

The other tokens are won't be supported, or just not discussed yet? If latter, how about that taishi-san works on typography related tokens and I work on the others?

poteboy:
@yukukotani 
That sounds like a great idea, go ahead with it🙏. Regarding the `spacing` token, I've been considering making it so that, like Chakra, the default is 0.25rem and setting `p={4}` would make it 1rem. I think it would be good to be able to set the default for this. I've created `packages/system/src/toCSS.ts` with this in mind, so please check it out!

yukukotani:
@poteboy Okay!

>  I've been considering making it so that, like Chakra, the default is 0.25rem and setting p={4} would make it 1rem. 

Do you mean that Kuma UI provides default tokens? I'm little confused because the doc says that [Kuma UI doesn't come with a default theme](https://www.kuma-ui.com/docs/Theme/CustomizingTheme#:~:text=it%20doesn%27t%20come%20with%20a%20default%20theme).

Maybe that means just components are headless?

yukukotani:
Finally, all the tokens listed have been supported. Let's close this

poteboy:
Thank you so much 🐻‍❄️👋

PR #289, opened 2023-08-19
Add support for size token

yukukotani:
SSIA #257

### DIFF
--- a/packages/system/src/layout.test.ts
+++ b/packages/system/src/layout.test.ts
@@ -1,0 +1,60 @@
+import { theme } from "@kuma-ui/sheet";
+import { LayoutProps } from ".";
+import { grid } from "./grid";
+import { beforeAll, describe, expect, test } from "vitest";
+import { layout } from "./layout";
+
+describe("layout utility function", () => {
+  beforeAll(() => {
+    theme.setUserTheme({
+      sizes: {
+        sm: "1rem",
+      },
+    });
+  });
+
+  // Arrange
+  const testCases: Array<[LayoutProps, string, string]> = [
+    [{ width: 8 }, "width: 8px;", ""],
+    [{ minWidth: "12px" }, "min-width: 12px;", ""],
+    [{ height: "2em" }, "height: 2em;", ""],
+    [{ maxHeight: "sm" }, "max-height: 1rem;", ""],
+    [{ display: "flex" }, "display: flex;", ""],
+    [{ display: "sm" }, "display: sm;", ""], // size token won't be applied
+  ];
+
+  test.each(testCases)(
+    "should return the correct CSS styles for the given LayoutProps",
+    (props, expectedBaseStyles, expectedMediaStyles) => {
+      // Act
+      const { base, media } = layout(props);
+      const mediaString = Object.entries(media)
+        .map(
+          ([breakpoint, css]) => `@media (min-width: ${breakpoint}) {${css}}`
+        )
+        .join("");
+
+      // Assert
+      expect(base.replace(/\s/g, "")).toBe(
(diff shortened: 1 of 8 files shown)


website/src/pages/docs/Theme/ThemeTokens.mdx
```diff
@@ -110,6 +110,20 @@ const theme = createTheme({
 });
 ```
 
+## Sizes
+
+Determine the size ob objects such as `width` and `height`.
```
poteboy, line 115: typo?
```suggestion
Determine the size of objects such as `width` and `height`.
```

yukukotani: oops, nice catch!

packages/system/src/layout.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    // Arrange
+    const invalidProps = {
+      invalid: 8,
+      invalid2: "8px",
+    };
+
+    // Act
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-explicit-any -- FIXME
+    const { base, media } = grid(invalidProps as any);
+    expect(base).toBe("");
```
poteboy, line 57: Nits: missing // Assert comment

yukukotani: [69c963f](https://github.com/kuma-ui/kuma-ui/pull/289/commits/69c963fdcf1807aae2a3a89623600fbbb6b25537)

packages/system/src/valueConverters.ts
```diff
@@ -7,3 +7,8 @@ export const spaceConverter: ValueConverter = (v) => {
   const userTheme = theme.getUserTheme();
   return userTheme.spacings?.[v] || toCssUnit(v);
 };
+
+export const sizeConverter: ValueConverter = (v) => {
+  const userTheme = theme.getUserTheme();
+  return userTheme.sizes?.[v] || toCssUnit(v);
```
poteboy, line 13: I'm not very familiar with ESLint, but did this line not throw errors like `eslint@typescript-eslint/no-unsafe-return` or `eslint@typescript-eslint/no-unsafe-member-access` in your local env? Since the CI isn't failing, I'm wondering if it's just an issue on my end?

yukukotani: @poteboy No error in my machine. Maybe you missed to run `pnpm build`?

poteboy: I re-ran `pnpm build` but still encountered the error. 
Interestingly, the error disappeared when I turned off the [VSCode plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).

poteboy (approved):

poteboy:
**MEMO**:
Ultimately, to avoid having the return type of `userTheme.sizes?.[v]` as `any`, we need to set the return type of the following section in `@kuma-ui/sheet` to `StyledProps`. However, doing this right now would create a circular dependency, making it unfeasible unless we fundamentally refactor the package structure.
```ts
getVariants(componentName: ComponentName):
| {
// eslint-disable-next-line @typescript-eslint/no-explicit-any -- FIXME
baseStyle?: any;
variants?:
| {
// eslint-disable-next-line @typescript-eslint/no-explicit-any -- FIXME
[key: string]: any;
}
| undefined;
}
| undefined {
return this._userTheme.components?.[componentName] || {};
}
```
So, let's keep it as it is for now.

yukukotani:
hmm I don't think the problem is that point but yes, let's investigate it later.

PR #274, opened 2023-08-06
Add support for spacing token

yukukotani:
Add `spacings` theme token for properties such as `margin`, `padding`, and more.

Mostly similar to #272 but I've introduced `ValueConverter` to simplify the transforming logic. So I'd like to get review of @taishinaritomi.

PR #290, opened 2023-08-19
deploy real example project to codesandbox

yukukotani:
Deploys [our vite example](https://github.com/kuma-ui/kuma-ui/tree/main/example/vite) from codesandbox CI instead of general react example to ease the effort of checking the behavior.

#### Before

https://github.com/kuma-ui/kuma-ui/pull/274#issuecomment-1676238005

#### After

https://github.com/kuma-ui/kuma-ui/pull/290#issuecomment-1684899949

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -2,6 +2,6 @@
   "installCommand": "codesandbox:insall",
   "buildCommand": "codesandbox:build",
   "packages": ["packages/*"],
-  "sandboxes": ["/example/webpack"],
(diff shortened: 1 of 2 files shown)


poteboy (approved):
Looks good, but since @kotarella1110  initially introduced the `codesandbox` to support the [Waku](https://github.com/dai-shi/waku) RSC framework, I'd like @kotarella1110 to review for validation.

kotarella1110 (approved):
Thanks! LGTM!

PR #292, opened 2023-08-19
fix/kuma.config can hmr

nojiritakeshi:
This PR is intended to reflect the contents of kuma.config.ts on the screen with hmr.


packages/system/src/generator.ts
```diff
@@ -21,7 +21,21 @@ export class StyleGenerator {
...
+        const propKey = propValue.split(".")[0] as StyleThemeKeyType;
+        if (userTheme[propKey] !== undefined) {
+          for (const key in userTheme[propKey]) {
+            if (propValue.trim() === key) {
+              styledProps[propName.trim()] = userTheme[propKey]![key];
+              break;
+            }
+          }
+        }
+      } else if (isStyledProp(propName.trim())) {
```
kotarella1110, line 38: Themes like `fonts` and `spacings` are increasing, and we plan to add various themes in the future as well.
As there are cases where `converter` differ for each theme token, it's not practical to process theme values all together here.
Could you please revert the changes to this code? :pray:

packages/system/src/color.ts
```diff
@@ -55,23 +54,7 @@ export const color = (props: ColorProps): ResponsiveStyle => {
...
-          return value;
-        };
-      } else {
-        converter = (v) => v;
-      }
-      const responsiveStyles = applyResponsiveStyles(
-        property,
-        cssValue,
-        converter
-      );
```
kotarella1110, line 74: For the same reasons as the previous comment, I would like you to revert the changes to this code as well.

kotarella1110:
I have tested it locally, but it seems that changes to the theme are not being applied. Please check and confirm.

https://github.com/kuma-ui/kuma-ui/assets/12913947/6fcfa801-4562-4bd5-ad4d-2a7cb407083a

nojiritakeshi:
@kotarella1110 
> I have tested it locally, but it seems that changes to the theme are not being applied. Please check and confirm.

To reproduce it locally, I think you need to specify it in the relative path of "transform,compile, StyleGenerator".

packages/system/src/generator.ts
kotarella1110: Sorry, please wait. I checked your Discord message. I will review again after confirming.

kotarella1110: This code doesn't seem to need to be reverted, so I resolve conversation.

packages/system/src/color.ts
kotarella1110, line 74: I think that [font](https://github.com/kuma-ui/kuma-ui/blob/main/packages/system/src/font.ts) etc. can also be changed.

kotarella1110:
> To reproduce it locally, I think you need to specify it in the relative path of "transform,compile, StyleGenerator".

Could you provide more detailed explanation about this a bit further?
If possible, it would be appreciated if you could also provide the code that runs on CodeSandbox.

nojiritakeshi:
@kotarella1110 
> Could you provide more detailed explanation about this a bit further?
If possible, it would be appreciated if you could also provide the code that runs on CodeSandbox.

Can you check it out on Github?
https://github.com/nojiritakeshi/kuma-ui/commit/154de4ffae309d8adc22693411502368b7c6d0f8

packages/system/src/color.ts
nojiritakeshi: I see that other things like fontSize have also been added!
I'll check it out and make some additional corrections.

kotarella1110:
@nojiritakeshi 

> Can you check it out on Github?

Thanks for providing the code to reproduce it.
I tried changing to a relative path, but the theme changes were not reflected in the browser.
Even the CodeSandbox CI example doesn't seem to work. You can check the CodeSandbox CI example at the following link.

スクリーンショット 2023-08-25 21 48 40

nojiritakeshi:
@kotarella1110 
hmm...
Can I confirm on discord?

kotarella1110:
@nojiritakeshi Of course 👍

packages/vite/src/index.ts
```diff
@@ -91,6 +89,9 @@ export default function kumaUI(): Plugin {
       server.ws.send({ type: "full-reload" });
     },
     configResolved(config) {
+      if (config.define && Object.keys(config.define).length > 0) {
```
poteboy, line 92: 👍

nojiritakeshi:
@kotarella1110 @poteboy 
Committed with modifications.
Please check back when you are available.

packages/compiler/src/extractor/extract.ts
```diff
@@ -49,7 +36,21 @@ export const extractProps = (
...
+      const propKey = propValue.split(".")[0] as Tokens;
+      if (userTheme[propKey] !== undefined) {
+        for (const key in userTheme[propKey]) {
+          if (propValue.trim() === key) {
+            styledProps[propName.trim()] = userTheme[propKey]![key];
+            break;
+          }
+        }
+      }
+    } else if (isStyledProp(propName.trim())) {
```
kotarella1110, line 53: @nojiritakeshi 
With this change, the HMR now works 🎉
However, I don't think it will work with Dynamic Props because the process has been moved from the system to the compiler.

```tsx
<Box p={12} color={() => "colors.blue.light"}>dynamic</Box>
```

I think it would be better to move this process to webpack-plugin and vite-plugin.

kotarella1110: I think this process needs to be implemented on the runtime side as well.
See the following line:

https://github.com/kuma-ui/kuma-ui/blob/7ebe9ed1120b9cd5cf451c1de124fab0abd49c6a/packages/core/src/components/Box/react/utils.ts#L30

kotarella1110 (approved):
@nojiritakeshi Great work 🏆 Thanks!

packages/system/src/column.ts
```diff
@@ -37,8 +36,8 @@ const columnMappings: Record<ColumnKeys, string> = {
   columns: "columns",
 };
 
-const converters: Partial<Record<ColumnKeys, ValueConverter>> = {
-  columnGap: spaceConverter,
+const converters: Partial<Record<ColumnKeys, typeof toCssUnit>> = {
```
kotarella1110, line 39: Is it possible to move the `ValueConverter` type to `system/src/types.ts` instead of removing it? I thought we should use that type here.

packages/system/src/border.test.ts
```diff
@@ -1,16 +1,7 @@
...
+import { describe, expect, test } from "vitest";
 
 describe("border utility function", () => {
-  beforeAll(() => {
-    theme.setUserTheme({
-      radii: {
-        sm: "1rem",
-      },
-    });
-  });
```
kotarella1110, line 12: Is it possible to keep the theme test instead of removing it?

packages/system/src/space.test.ts
```diff
@@ -1,16 +1,7 @@
...
+import { describe, expect, test } from "vitest";
 
 describe("space utility function", () => {
-  beforeAll(() => {
-    theme.setUserTheme({
-      spacings: {
-        sm: "1rem",
-      },
-    });
-  });
```
kotarella1110, line 12: Same as above comment.

kotarella1110 (commented):
@nojiritakeshi I am sorry that I approved your code, but I commented on it and you need to fix it. Once the corrections are complete, I will approve and merge again.

yukukotani (commented):
Can't wait to get this landed!

packages/system/src/generator.ts
```diff
@@ -21,7 +21,21 @@ export class StyleGenerator {
...
+      ) {
+        const userTheme = theme.getUserTheme();
+        const propKey = propValue.split(".")[0] as Tokens;
+        if (userTheme[propKey] !== undefined) {
+          for (const key in userTheme[propKey]) {
+            if (propValue.trim() === key) {
+              styledProps[propName.trim()] = userTheme[propKey]![key];
+              break;
+            }
+          }
```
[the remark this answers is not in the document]
poteboy: The reason for needing `trim()` on propValue is actually because `getFullText` is used instead of `getText` within the `collector` (My fault 🤦). I believe that fixing those area might eliminate the need for this adjustment. 
Whether or not this will be addressed in this PR is up to @nojiritakeshi  though

packages/system/src/border.test.ts
nojiritakeshi: @kotarella1110 
These processes would actually take place after the user theme has already been converted.
Therefore, I believe that testing the theme is unnecessary for this test

kotarella1110: @nojiritakeshi You are right. I think this line can be deleted.
However, you should keep the following test case!

https://github.com/kuma-ui/kuma-ui/pull/292/files#diff-a474cb7251af443b69bc46e96a3ce239067c318dd9f38a86cca45ef2b8020c6cL78

nojiritakeshi: @kotarella1110 
Is it correct that the code that should be left behind is the following?
`[{ borderRadius: "sm" }, "border-radius: 1rem;", ""],`

I think that the value passed as "sm" cannot be 1rem at the time this process is performed.

packages/system/src/generator.ts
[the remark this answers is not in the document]
nojiritakeshi: @yukukotani @poteboy 
I'll try to fix it and include it in this PR, but I'll separate the commits.

packages/system/src/border.test.ts
kotarella1110: @nojiritakeshi It means that I want to keep the test case of borderRadius itself as follows.

```
[{ borderRadius: "1rem" }, "border-radius: 1rem;", ""],
```

packages/system/src/generator.ts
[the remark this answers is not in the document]
nojiritakeshi: @poteboy 
I removed all the trim()'s that I thought were unnecessary, but all the existing tests were fine.
I also checked the actual operation of the system and it was fine as far as I was concerned!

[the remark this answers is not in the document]
poteboy: Thank you 🍑

kotarella1110 (approved):
Thank you for your hard work 🎉🎉🎉

poteboy (approved):

PR #293, opened 2023-08-19
Add support for z-index token

yukukotani:
SSIA #257


packages/core/src/theme.test.ts
```diff
@@ -115,6 +115,57 @@ describe("createTheme", () => {
     }>();
   });
 
+  test("should convert radius theme object to expected format", () => {
```
yukukotani, line 118: Forgot this in previous PR

poteboy (approved):
💜

PR #297, opened 2023-08-23
Remove regex lookbehinds to support Safari 16.3 or older

myzkyy:
Lookbehind assertion is not supported by Safari 16.3 (Released on Jan 23, 2023) or older. It causes errors: `SyntaxError: Invalid regular expression: invalid group specifier name`.

cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion#browser_compatibility

This PR removes all of them.

### DIFF
--- a/.changeset/mighty-coats-breathe.md
+++ b/.changeset/mighty-coats-breathe.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/sheet": minor
+---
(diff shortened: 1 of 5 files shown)


myzkyy:
Bad branch name. Recreated #297.

PR #298, opened 2023-08-23
Remove regex lookbehinds to support Safari 16.3 or older 

myzkyy:
Lookbehind assertion is not supported by Safari 16.3 (Released on Jan 23, 2023) or older. It causes errors: `SyntaxError: Invalid regular expression: invalid group specifier name.`

cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion#browser_compatibility

This PR removes all of them.

### DIFF
--- a/.changeset/mighty-coats-breathe.md
+++ b/.changeset/mighty-coats-breathe.md
@@ -2,4 +2,4 @@
 "@kuma-ui/sheet": minor
 ---
 
-Remove regex lookbehinds to support Safari <= 16.3
+Remove regex lookbehinds to support Safari <= 16.3 and some refactoring
--- a/.changeset/mighty-coats-breathe.md
+++ b/.changeset/mighty-coats-breathe.md
@@ -1,5 +1,5 @@
 ---
-"@kuma-ui/sheet": minor
+"@kuma-ui/sheet": patch
 ---
 
 Remove regex lookbehinds to support Safari <= 16.3 and some refactoring
--- a/.changeset/mighty-coats-breathe.md
+++ b/.changeset/mighty-coats-breathe.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/sheet": minor
+---
+
+Remove regex lookbehinds to support Safari <= 16.3
--- a/packages/sheet/src/sheet.ts
+++ b/packages/sheet/src/sheet.ts
@@ -1,6 +1,9 @@
 import { theme } from ".";
 import { generateHash } from "./hash";
-import { cssPropertyRegex, removeSpacesExceptInPropertiesRegex } from "./regex";
+import {
+  removeSpacesAroundCssPropertyValues,
+  removeSpacesExceptInProperties,
+} from "./regex";
 import { compile, serialize, stringify, Element } from "stylis";
 
 // to avoid cyclic dependency, we declare an exact same type declared in @kuma-ui/system
@@ -65,8 +68,8 @@ export class Sheet {
   }
 
   private _addBaseRule(className: string, css: string) {
-    css = css.replace(cssPropertyRegex, "$1$2");
-    this.base.push(`.${className}{${css}}`);
+    const minifiedCss = removeSpacesAroundCssPropertyValues(css);
+    this.base.push(`.${className}{${minifiedCss}}`);
   }
 
   private _addMediaRule(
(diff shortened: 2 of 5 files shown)


poteboy:
@kotarella1110  @taishinaritomi 
Hey, I want to hear your opinion on the approach this PR took. This kinda led me to revisit the core purpose of this regex:
- Most libraries let bundler plugins handle CSS minification.
- We create some CSS at runtime (Therefore we need some minification logic like this regex).
- But is it our job to also minify this? Especially when tools like [matthiasmullie/minify](https://github.com/matthiasmullie/minify) exist.
- On the other hand, keeping external dependencies low is good for bundle size and maintenance.

It's a balance. Should we keep this minify process or let other tools handle it?

myzkyy:
@poteboy Thank you for your consideration. I wonder what the meaning of runtime minification is. Minification is for traffic efficiency, isn't it?

By the way, as you may already know, this error blocks running scripts on the website. For example, In https://www.kuma-ui.com/ users can't open the hamburger menu if they use Safari 16.3. I insist this is a critical problem for production use.

kotarella1110:
If it's a reduction logic of the extent to remove spaces, I believe it's fine to keep it. If you want to further reduce it, the user should use a minify tool on their side.

packages/sheet/src/sheet.ts
```diff
@@ -74,11 +74,11 @@ export class Sheet {
...
     breakpoint: string
   ): void {
-    css = css.replace(cssPropertyRegex, "");
+    css = css.replace(cssPropertyRegex, "$1$2");
     const mediaCss =
       `@media (min-width: ${breakpoint}) { .${className} { ${css} } }`.replace(
         removeSpacesExceptInPropertiesRegex,
-        ""
+        "$1$2$3"
       );
```
kotarella1110, line 82: @myzkyy Is it possible to extract this whitespace removal logic as a function?

example:

```ts
const removeSpacesExceptInProperties = (str: string) => str.replace(removeSpacesExceptInPropertiesRegex, "$1$2$3")
```

myzkyy: @kotarella1110 Done.

poteboy:
duplicate changesets found

myzkyy:
> duplicate changesets found

Merged.

kotarella1110 (approved):
Thanks! LGTM 🎉

packages/sheet/src/sheet.ts
```diff
@@ -65,32 +68,28 @@ export class Sheet {
...
   private _addPseudoRule(
     className: string,
     pseudo: SystemStyle["pseudo"][number]
   ) {
-    const css = pseudo.base.replace(cssPropertyRegex, "");
-    const pseudoCss = `.${className}${pseudo.key} { ${css} }`.replace(
-      removeSpacesExceptInPropertiesRegex,
-      ""
+    const pseudoCss = removeSpacesExceptInProperties(
+      `.${className}${pseudo.key} { ${pseudo.base} }`
```
myzkyy, line 92: I noticed unintended behavior change. I'll fix here.

myzkyy: `removeSpacesExceptInProperties` contains all of `removeSpacesAroundCssPropertyValues` regexp, so this diff causes no behavior change actually.

myzkyy: Anyway I added commit c34ace78b0dc028299f022c6d9c5eb41d10d377c

myzkyy: @kotarella1110 Could you review change c34ace78b0dc028299f022c6d9c5eb41d10d377c?
I added some refactoring, too.

kotarella1110: thanks! will review 👍

.changeset/mighty-coats-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kuma-ui/sheet": minor
```
kotarella1110, line 2: @myzkyy I think the patch version is better because this is a bugfix.

myzkyy: @kotarella1110 Fixed 477c3ffa64ea19d7ad63e94e2eeb8bad12e18b82

kotarella1110 (approved):
looks good 👍

PR #300, opened 2023-08-24
publish dev snapshot on main branch commit

yukukotani:
Introduces snapshot release, which is being published on every commit on main branch.

Honestly I can't test this but it should work in theory.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,3 +29,13 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish dev snapshot
+        if: steps.changesets.outputs.published != 'true'
+        run: |
+          git checkout main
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -29,3 +29,13 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish dev snapshot
+        if: steps.changesets.outputs.published != 'true'
+        run: |
+          git checkout main
+          pnpm changeset version --snapshot dev
+          pnpm changeset publish --tag dev
```
yukukotani, line 38: changesets/action has no support of snapshots (https://github.com/changesets/action/issues/150) so doing manually.

poteboy (approved):
Thanks! Let's see if this works 👍

yukukotani:
Oops it failed. Taking a look now

https://github.com/kuma-ui/kuma-ui/actions/runs/5965772915/job/16183953680

PR #304, opened 2023-08-25
Improve autocompletion experience of `baseStyle` and `variants`

yukukotani:
Currently we can't get any completion in `baseStyle` and `variants` of `kuma.config.js`.

With this PR, we can get completion of property names and values. Still cannot autocomplete custom token as it causes circular reference but it's better experience than `any` type.


kotarella1110 (approved):
Nice improvement 🎉

Issue #303, opened 2023-08-25
Allow to specify `defaultProps` on component themes


yukukotani:
For more flexibility in customising headless components, introduce the `defaultProps' option to component theme.

The priority would be like:

1. The properties given in JSX
2. The `defaultProps` option in component theme (new)
3. The default properties for each component such as `cursor: pointer` for Button

### Real example

```js
const theme = createTheme({
  components: {
    Button: {
      baseStyle: {
        borderRadius: '4px', // baseStyle specifies only the skeleton of component
      },
      variants: {
        primary: {
          color: 'white',
          bgColor: 'blue',
        },
        secondary: {
          border: 'solid 1px black',
          color: 'blue',
        },
      },
      defaultProps: {
        variant: 'primary', // specify default variant
        cursor: 'grab', // override component's default prop
      },
    },
  },
});
```

poteboy:
This would be dope ✌️

yukukotani:
Good, let me take this task

yukukotani:
#306

PR #306, opened 2023-08-26
Introduce `defaultProps` in component theme

yukukotani:
#303

Introduce `defaultProps` to allow users to define default properties for each component.

### Example

```js
const theme = createTheme({
  components: {
    Text: {
      baseStyle: {
        textDecoration: "underline",
      },
      variants: {
        primary: {
          color: "green",
        },
        secondary: {
          color: "blue",
        },
      },
      defaultProps: {
        variant: "primary",
        textDecoration: "line-through",
        fontSize: "16px",
      },
    },
  },
});
```

Then the CSS for `<Text />` will be like `{ color: green;text-decoration: line-through;font-size: 16px; }`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "scripts": {
     "bootstrap": "pnpm install",
     "build": "turbo run build",
+    "build:packages": "turbo run build --filter @kuma-ui/*",
     "test": "turbo run test",
     "lint": "eslint './packages/**/*.{js,ts,jsx,tsx}' --report-unused-disable-directives --max-warnings 0",
     "lint:fix": "eslint --fix './packages/**/*.{js,ts,jsx,tsx}' --report-unused-disable-directives --max-warnings 0",
--- a/packages/babel-plugin/src/__test__/component.test.ts
+++ b/packages/babel-plugin/src/__test__/component.test.ts
@@ -2,7 +2,7 @@ import { theme } from "@kuma-ui/sheet";
 import { babelTransform, getExpectSnapshot } from "./testUtils";
 
 describe("headless component", () => {
-  test("using default props should match snapshot", () => {
+  beforeAll(() => {
     theme.setUserTheme({
       components: {
         Text: {
@@ -25,11 +25,28 @@ describe("headless component", () => {
         },
       },
     });
+  });
+
+  test("using default props should match snapshot", () => {
     // Arrange
     const inputCode = `
         import { Text } from '@kuma-ui/core'
         function App() {
           return <Text />
+        }
+      `;
+    // Act
+    const result = babelTransform(inputCode);
+    // Assert
+    expect(getExpectSnapshot(result)).toMatchSnapshot();
+  });
+
+  test("using default props with inline props should match snapshot", () => {
+    // Arrange
+    const inputCode = `
+        import { Text } from '@kuma-ui/core'
+        function App() {
+          return <Text variant="secondary" fontSize="12px" />
         }
       `;
     // Act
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
   "scripts": {
     "bootstrap": "pnpm install",
     "build": "turbo run build",
+    "build:packages": "turbo run build --filter @kuma-ui/*",
```
yukukotani, line 36: Add faster build script since we don't need to build examples and website for most tasks.

yukukotani:
@poteboy I have two things that need your opinion.

1. How can I add tests for this? IMO we need some refactors of the compiler to write unit test for this change.
2. `defaultProps` and `baseStyle` may seem to overlap a little in concept. Shoud we merge them?

poteboy:
@yukukotani 
> defaultProps and baseStyle may seem to overlap a little in concept. Shoud we merge them?

What do you mean by 'merge' in this context? Are you referring to merging them internally or do you mean integrating the functionalities themselves? [Chakra UI](https://chakra-ui.com/docs/styled-system/customize-theme#theme-extension-withdefaultprops) also has both baseStyle and defaultProps, so..

yukukotani:
@poteboy I mean latter, declaring both base StyleProps and variants in same object.

poteboy:
@yukukotani 
Indeed, I see your point about not needing both. However, this would be a breaking change for Kuma's API, requiring announcements and a migration guide. I'd like to discuss this in our core team meeting first. Hold tight 🙏
On a personal note, would you be interested in joining one of our core team meetings?

packages/core/src/theme.ts
```diff
@@ -57,3 +58,15 @@ export function createTheme<const T extends ThemeInput>(
     components: components,
   } as unknown as ThemeResult<T>;
 }
+
+type ComponentThemeInput<Variants extends { [key: string]: StyledProps }> = {
+  baseStyle?: StyledProps;
+  variants?: Variants;
+  defaultProps?: StyledProps & { variant?: keyof Variants };
```
poteboy, line 65: There are use cases where you'd want to pass HTML attributes like WAI-ARIA, so having a type like `Record<string, any>` might be appropriate?

yukukotani:
@poteboy Sure, let's not rush into the decision!

I have both plans that require migration and vice versa, so wanna introduce those to you. It'll be so excited to cooperate more closely with brilliant core members :+1:

packages/core/src/theme.ts
yukukotani: True! Also `StyleProps` was unnecessary since `baseStyle` handle them. [a68aa1f](https://github.com/kuma-ui/kuma-ui/pull/306/commits/a68aa1f036895e4c5cb4465a62c3a1bc730882ff)

packages/babel-plugin/src/__test__/component.test.ts
yukukotani, line 1: We need some refactoring to add test to compiler package so adding tests in babel-plugin instead, for now.

packages/compiler/src/extractor/extract.ts
```diff
@@ -41,12 +41,14 @@ export const extractProps = (
     ...(variant?.baseStyle as Record<string, string>),
   };
 
-  const defaultProps = componentDefaultProps(componentName);
+  const systemDefaultProps = componentDefaultProps(componentName);
+  const userDefaultProps = variant?.defaultProps;
```
poteboy, line 45: I had previously declared a variable named `componentVariantProps` to accumulate variant-related props. However, you introduced a new variable `userDefaultProps` and did not utilize `componentVariantProps`, which I believe is perfectly fine. I would kindly suggest renaming `componentVariantProps` to something more descriptive like `baseStyle` to better align with its intended use case.

taishinaritomi (approved):

PR #309, opened 2023-08-26
fix(vite): fix incorrect exports.types path

apple-yagi:
## Summary
This Pull Request addresses the issue of an incorrect `exports.types` path specified in the `packages.json` file.

## Background
The current incorrect path specification is causing issues with referencing TypeScript type definition files. This fix ensures that the correct type definition files are accessible.
スクリーンショット 2023-08-26 23 26 19

## Changes Made
The correct `exports.types` path has been updated.


poteboy (approved):
I did not noticed that, thanks 😅

PR #317, opened 2023-09-01
Run CI with node v18

yukukotani:
Use current LTS Node.js v18 to make build faster and ensure kuma works on v18

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node_version: [16]
+        node_version: [16, 18]
         os: [ubuntu-latest, windows-latest]
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -5,18 +5,15 @@ on:
       - main
 jobs:
   release:
-    name: Node ${{ matrix.node_version }} releasing
+    name: Release
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node_version: [16]
```
yukukotani, line 12: Release job should not be matrix

.github/workflows/main.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node_version: [16]
+        node_version: [16, 18]
```
yukukotani, line 19: Running test on both v16 and v18 to make sure kuma keep supporting v16

poteboy (approved):

PR #323, opened 2023-09-06
Documenting the Transition from `baseStyle` to `defaultProps`

poteboy:
Follow-up to PR #306


yukukotani (approved):
Oops I greped with `baseStyle` so missed them... thanks!

PR #324, opened 2023-09-06
improve createTheme type

kotarella1110:
The type of theme property values was changed in https://github.com/kuma-ui/kuma-ui/pull/306 from being returned as primitive types to being returned as literal types.


yukukotani (approved):

poteboy:
Life savior

taishinaritomi (approved):

Issue #318, opened 2023-09-03
Allow using theme tokens in css and styled


seagullua:
Currently, using theme tokens to style third-party components is impossible.

It would be nice to have possibility to write

```
className={css`
                      color: theme("colors.primary");
                    `}
```

And have `theme("colors.primary")` replaced with real value at the build

Alternatively, it kuma may export all theme tokens as CSS variables so you can write `var(--kuma-colors-primary)`.

poteboy:
@seagullua 
Absolutely, but I am pondering the API design. I would think it is easier to just write `colors.primary`. What do you say?

seagullua:
@poteboy, thanks for a quick reply. 

I think the good idea will be to stick with something like theme(...) or token(...)

1. It will be easier to parse and harder to make the wrong replacement. For example, right now if I define breakpoint: sm. It replaces `@media (... small)` with `@media (... 450pxall)`. Some keywords will solve lots of such issues.
2. Right now, Webstorm will show an error for `colors.xxxx ` but not for keywords.

<img width="847" alt="Screenshot 2023-09-04 at 9 57 29 AM" src="https://github.com/kuma-ui/kuma-ui/assets/759038/f5a4b6b6-bc51-4ac1-ae25-3e316afad982">

poteboy:
> Webstorm will show an error for colors.xxxx  but not for keywords

Thank you for your suggestion. I hadn't considered the potential issues with editors indicating errors for `colors.xxxx`, but it's certainly something worth keeping in mind.

seagullua:
@poteboy Do you think the implementation can take long? 

I was checking the source code, and it looks like it can be straightforward. You need a regular expression to catch all theme(...) in the code and match the known keys.

I would have implemented it, but I did not figure out how to debug babel preprocessors.

poteboy:
@seagullua 

I suspect that the `theme(...)` function might currently be recognized as a CSS function in your editor. Considering this, I have a concern that using a generic name like theme for this function might clash with future web standards. So, how about we opt for a more distinctive syntax such as `t(...)` to avoid potential naming conflicts?

Regarding debugging, it seems like the viable approach for now is to make modifications in the packages, build it, and then verify the behavior within a project in the playground directory. You might find it easier to check the AST transformations in the browser using the [vite-plugin-inspect](https://github.com/antfu/vite-plugin-inspect) within the `playground/vite project`.

seagullua:
[...]

PR #327, opened 2023-09-06
#318 Theme tokens support

seagullua:
As per https://github.com/kuma-ui/kuma-ui/issues/318 I have added tokens placeholders you can use in css

```
css`
   color: t("color.primary");
   background-color: t("c.primary");
   
  @media (max-width: t("b.sm")) {
     margin-top: t(2);
  }
`
```

Will generate CSS

```
   color: red;
   background-color: red;
   
  @media (max-width: 800px) {
     margin-top: 16px;
  }
```

What is supported
- `t("<token>.<key>")` any token specified in kuma.config.js
- ~`t("c.<key>")` and `t("b.<key>")` shortcuts for colors and breakpoints~
- ~`t(2)` scaling factor like https://mui.com/material-ui/customization/spacing/ . Right now, it is hardcoded as 8px, but it is supper easy to add config to kuma.config~

### DIFF
--- a/packages/sheet/src/placeholders.test.ts
+++ b/packages/sheet/src/placeholders.test.ts
@@ -1,0 +1,262 @@
+import {
+  applyPlaceholders,
+  applySpacingScalingFactor,
+  createPlaceholders,
+  applyT,
+  applyArrayT,
+} from "./placeholders";
+import { describe, expect, it } from "vitest";
+import { UserTheme } from "./theme";
+
+describe("Theme utility functions", () => {
+  describe("applyPlaceholders", () => {
+    it("replaces valid placeholders", () => {
+      const input = "Color is t('c.primary') and breakpoint is t('b.mobile')";
+      const placeholders = {
+        "c.primary": "#FF0000",
+        "b.mobile": "480px",
+      };
+      expect(applyPlaceholders(input, placeholders)).toBe(
+        "Color is #FF0000 and breakpoint is 480px"
+      );
+    });
+
+    it("ignores invalid placeholders", () => {
+      const input = "Color is t('c.primary') and something is t('invalid.key')";
+      const placeholders = {
+        "c.primary": "#FF0000",
+      };
+      expect(applyPlaceholders(input, placeholders)).toBe(
+        "Color is #FF0000 and something is t('invalid.key')"
+      );
+    });
+  });
+
+  describe("applySpacingScalingFactor", () => {
+    it("replaces integers with scaled px value", () => {
+      const input = "Margin is t(2)";
+      expect(applySpacingScalingFactor(input, 8)).toBe("Margin is 16px");
+    });
+
+    it("replaces decimals with scaled px value", () => {
+      const input = "Padding is t(2.5)";
+      expect(applySpacingScalingFactor(input, 8)).toBe("Padding is 20px");
+    });
+
+    it("ignores non-number values", () => {
+      const input = "Something is t(notANumber)";
+      expect(applySpacingScalingFactor(input, 8)).toBe(
+        "Something is t(notANumber)"
+      );
+    });
+  });
+
+  describe("createPlaceholders", () => {
+    it("creates placeholders for valid theme values", () => {
+      const theme: Partial<UserTheme> = {
+        colors: {
+          primary: "#FF0000",
+        },
+        breakpoints: {
+          mobile: "480px",
+        },
+      };
+
+      expect(createPlaceholders(theme)).toEqual({
+        "colors.primary": "#FF0000",
+        "c.primary": "#FF0000",
+        "breakpoints.mobile": "480px",
+        "b.mobile": "480px",
+      });
+    });
+
+    it("ignores invalid or non-existent theme values", () => {
+      const theme: Partial<UserTheme> = {
+        colors: {
+          primary: "#FF0000",
+        },
+      };
+
+      expect(createPlaceholders(theme)).toEqual({
+        "colors.primary": "#FF0000",
+        "c.primary": "#FF0000",
+      });
+    });
+  });
+
+  describe("applyPlaceholders edge cases", () => {
+    it("handles empty strings", () => {
+      const placeholders = { "c.primary": "#FF0000" };
+      expect(applyPlaceholders("", placeholders)).toBe("");
+    });
+
+    it("handles strings without placeholders", () => {
+      const placeholders = { "c.primary": "#FF0000" };
+      expect(applyPlaceholders("No placeholders here", placeholders)).toBe(
+        "No placeholders here"
+      );
+    });
+
+    it("replaces multiple occurrences of the same placeholder", () => {
+      const placeholders = { "c.primary": "#FF0000" };
+      expect(
+        applyPlaceholders(
+          "Color t('c.primary') and again t('c.primary')",
+          placeholders
+        )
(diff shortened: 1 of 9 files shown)


poteboy:
@seagullua 
Thank you for the PR!
The implementation of `t` as a shorthand for theme is a technical constraints, particularly to avoid potential conflicts with web standards. However, I find that the abbreviations `c`, `b`, and the use of numerical values are not immediately intuitive.

Moreover, Considering that our token system encompasses more than just colors and breakpoints, as outlined in our [documentation](https://www.kuma-ui.com/docs/Theme/ThemeTokens), I think this approach might cause conflicts in the future?

packages/sheet/src/placeholders.ts
```diff
@@ -0,0 +1,85 @@
+import { Tokens, UserTheme } from "./theme";
+
+const tokens: Tokens[] = [
```
poteboy, line 3: Considering future extensions to the `Tokens` type, it might be beneficial to define tokens with `as const` first, and then derive the `Tokens` type from it. This way, we can ensure type-level exhaustiveness checks automatically, making the code more robust against potential additions or modifications to the tokens array in the future.

```ts
export const tokens = [
  "colors",
  "fonts",
  "fontSizes",
  "fontWeights",
  "lineHeights",
  "letterSpacings",
  "spacings",
  "sizes",
  "radii",
  "zIndices",
  "breakpoints",
] as const;

export type Tokens = typeof tokens[number];
```

poteboy:
@seagullua 

I'm sorry, but at this moment, I'd like to keep the exception cases within `styled` or `css` limited to using `t()` only. Therefore, I'd prefer to discuss the shortcuts and spacing in a different section. At the very least, I want to keep it to that scope for this PR. Thank you for understanding. 🙏

seagullua:
@poteboy I have made changes as you asked
1. I have exported a list of placeholders
2. I have left only essential functionality t("colors.primary")

Please let me know if any other changes are required.

packages/sheet/src/placeholders.test.ts
```diff
@@ -0,0 +1,172 @@
+import { applyPlaceholders, createPlaceholders, applyT } from "./placeholders";
+import { describe, expect, it } from "vitest";
+import { UserTheme } from "./theme";
+
+describe("Theme utility functions", () => {
```
kotarella1110, line 5: please remove shortcuts c and b from test.

packages/sheet/src/sheet.test.ts
```diff
@@ -80,4 +81,37 @@ describe("Sheet class", () => {
...
+          sm: "1000px",
+        },
+        colors: {
+          primary: "red",
+          secondary: "blue",
+        },
+        zIndices: {
+          modal: "1000",
+        },
+      });
```
kotarella1110, line 98: There seems to be a mistake in this test because the return value of [createTheme](https://github.com/kuma-ui/kuma-ui/blob/e32721f5bdaa5e90b1a4f781483b25ededcc4c34/packages/core/src/theme.ts#L83-L101) API is passed as an argument to setUserTheme.

```suggestion
      theme.setUserTheme({
        breakpoints: {
          "breakpoints.sm": "1000px",
        },
        colors: {
          "colors.primary": "red",
          "colors.secondary": "blue",
        },
        zIndices: {
          "zIndices.modal": "1000",
        },
      });
```

packages/sheet/src/placeholders.ts
```diff
@@ -0,0 +1,40 @@
...
+  const result: Record<string, string> = {};
+
+  for (const token of tokens) {
+    const tokenValue = theme[token];
+    if (tokenValue) {
+      for (const key in tokenValue) {
+        result[`${token}.${key}`] = tokenValue[key]; // Add the token itself
+      }
+    }
+  }
```
kotarella1110, line 37: Considering that the return value of the createTheme API is passed, as mentioned in the comment above, the logic here needs to be fixed as follows.

```suggestion
  for (const token of tokens) {
    const tokenValue = theme[token];
    if (tokenValue) {
      for (const key in tokenValue) {
        result[key] = tokenValue[key]; // Add the token itself
      }
    }
  }
```

packages/sheet/src/sheet.test.ts
seagullua: The idea was to prevent the conflicts and to keep the existing structure of tokens.

So the placehorlders are build like <tokenType>.<tokenName>

The approach you are proposing will be less intuitive and more error prone as you can easily give different tokens the same name

kotarella1110: Hmmm, what does that mean ?
Themes defined in kuma.config.ts are exported as return values of createTheme API. The return value is passed as an argument to theme.setUserTheme. the return value of the createTheme API is converted to the structure I have suggested. Have you checked the following codes?
https://github.com/kuma-ui/kuma-ui/blob/e32721f5bdaa5e90b1a4f781483b25ededcc4c34/packages/core/src/theme.ts#L83-L101

kotarella1110: If kuma.config.ts was configured as follows,

```ts
// kuma.config.ts
import { createTheme } from "@kuma-ui/core";

const theme = createTheme({
  colors: {
    red: "red",
});
// theme: {
//   colors: {
//     "colors.red": "red",
//   },
// }

type UserTheme = typeof theme;

declare module "@kuma-ui/core" {
  export interface Theme extends UserTheme {}
}

export default theme;
```

Your code does not work with reference to theme as follows.

```ts
<Box className={css`color: t("colors.red")`}>test</Box>
```

Your code works by referencing theme as follows.

```ts
<Box className={css`color: t("colors.colors.red")`}>test</Box>
```

kotarella1110: Please see createTheme test: 
https://github.com/kuma-ui/kuma-ui/blob/e32721f5bdaa5e90b1a4f781483b25ededcc4c34/packages/core/src/theme.test.ts

packages/sheet/src/placeholders.ts
seagullua: @kotarella1110 @poteboy My bad. I did not figure out that when the theme is built the object is flattered.

1. I fixed that
2. Fixed a few more cases
3. Added new test

packages/sheet/src/placeholders.test.ts
seagullua: Done

packages/sheet/src/sheet.test.ts
```diff
@@ -33,6 +62,7 @@ describe("Sheet class", () => {
     expect(cssString).toContain(
       `.${className}{${removeSpacesAroundCssPropertyValues(style.base)}}`
     );
+    console.log(cssString);
```
kotarella1110, line 65: please remove console.log 🙏

seagullua: @kotarella1110 ready

kotarella1110 (commented):
@seagullua Thank you for addressing my feedback! The code looks good.

Would it be possible to add a test for the t function in the following snapshot tests?
Once that's added, I will approve.

- https://github.com/kuma-ui/kuma-ui/blob/main/packages/babel-plugin/src/__test__/css.test.ts
- https://github.com/kuma-ui/kuma-ui/blob/main/packages/babel-plugin/src/__test__/styled.test.ts

seagullua:
@kotarella1110 Ready. I have added tests to both.

PR #331, opened 2023-09-10
No type error when kuma-style is dynamically given to styled component

nojiritakeshi:
Originally, the following code works.
```ts
const Component = styled("p")`
  color: red;
`;

<Component fontSize={12}>Componentssss</Component>
```
The `font-size:12px` was applied at runtime, but there was a TypeScript type error in the code.

I created a PR in the hope that it may not be a fundamental solution to Styled's dynamic styling, but only to solve errors related to types.

### DIFF
--- a/.changeset/styled-component-type-error.md
+++ b/.changeset/styled-component-type-error.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/core": minor
+---
+
+No type error when kuma styles are given directly to Styled components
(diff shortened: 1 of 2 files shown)


poteboy:
@nojiritakeshi 
Components created using the `styled` API are not designed to accept style properties directly. I had previously removed the type but overlooked removing the runtime execution. I apologize for the confusion. Could you please close this PR? I appreciate your understanding. 🙇‍♂️

PR #334, opened 2023-09-14
fix: improve theme type for theme autocomplete to work

kotarella1110:
I have improved type autocompletion for Styled Props when configuring the following themes:

- radii
- sizes
- zIndices


poteboy (approved):
Thank you 🙇‍♂️

PR #342, opened 2023-09-23
styled-components syntax parity

### LINKED ISSUE #341: Support styled.div`...` syntax, like styled-components
Hi, this looks very promising - we're looking into kuma as a replacement for styled-components that works without overhead for react server components. We have hundreds of files and thousands of components to migrate, though, so the closer we can get to something with an API like styled-components, the better.

I noticed in the docs that this is how a div would be styled:

```ts
const MyDiv = styled('div')`
  color: blue;
`
```

Would you consider adding this syntax, to match styled-components and make our migration _much_ easier?

```ts
const MyDiv = styled.div`
  color: blue;
`
```

(as well as equivalents for the various HTML element types)

It could be a non-breaking, thin wrapper for the existing syntax. I'd be happy to contribute a PR for this if it'd be accepted.

mmkal:
Fixes #341

Adds a handler for `` styled.xxx` ` `` as an alternative to `` styled.xxx(` `) ``

Now exports `styled` as a `Proxy` which basically implements `styled.xxx` as `styled('xxx')`.

Also updates the type of `styled` so that it accepts either a string literal (corresponding to an HTML element), or an existing styled component (which is then extended, like in `styled-components`: https://styled-components.com/docs/basics#extending-styles)

Some tests for the compiler, and babel plugin were added, as well as some components using the new syntax added to the next-app-router example app.


poteboy:
@mmkal 

Thank you for your PR! Before reviewing the implementation, I'd like to request a couple of changes:
1. Instead of converting all instances of `styled('div')` in the documentation to `styled.div`, please make a note in the documentation that both syntaxes are supported.
2. We have a `__test__` directory in the compiler. Please add tests there to ensure that both syntaxes are supported.

packages/core/src/styled.ts
```diff
@@ -23,5 +21,127 @@ function styled<T extends keyof JSX.IntrinsicElements>(Component: T) {
   };
   return fn;
 }
+/**
+ * Kuma UI's `styled` API allows you to create styled React components using tagged template literals. This makes it a familiar and comfortable choice for developers who have worked with libraries like styled-components or Emotion.
+ */
+const styled = Object.assign(_styled, {
```
poteboy, line 27: While I understand the intent behind this implementation, I have some concerns regarding its maintainability, especially when new tags are introduced or if certain tags become deprecated. 
I'll give it some thought and see if we can come up with a more elegant solution tomorrow. (maybe utilizing `Proxy`?)

mmkal: Yeah, Proxy would work. Why don't I push a commit with Proxy to see how it looks.

poteboy:
@mmkal 
Thank you!
It's Nits, but lastly, I would like you to write a snapshot test in `packages/babel-plugin/src/__test__/styled.test.ts` to ensure that the newly introduced styled.div syntax correctly outputs CSS!

mmkal:
> ` [87bffe1](/kuma-ui/kuma-ui/pull/342/commits/87bffe130cddf5fa33fc03894c9ceeed8de9d4b5)`

@poteboy doess 87bffe1 look right?

Note that I did simplify the generated function a little bit - I got rid of the two `| ""` on the old and new `className`, since both were checked for truthiness already. And the indentation also changed for the sake of readability in this repo - but I am totally fine to revert to how it was if you prefer.

poteboy (approved):
@mmkal 
> Note that I did simplify the generated function a bit - I got rid of the two | "" on the old and new className, since both were checked for truthiness already. And the indentation also changed for the sake of readability in this repo - but I am totally fine to revert to how it was if you prefer.

LGTM! Thanks!
I agree, it's much more readable now. But I have a feeling that this will fail all snapshots generated by the styled API... Let's see how the tests will go.

poteboy:
> I have a feeling that this will fail all snapshots generated by the styled API

Oh It's already modified, nvm.

mmkal:
@poteboy I think this is ready now. Some questions re follow-ups I'd like to do:

1. I opened a draft on my fork against this branch to support `` const GreenButton = styled.button` background: green ` `` and then `` const GreenButtonRedText = styled(GreenButton)` color: red ` ``. I think this is a must have for anyone considering migrating from styled-components to kuma on a non-trivial existing project: https://github.com/mmkal/kuma-ui/pull/1. Once the current PR is merged I'll move that one to this repo, but feel free to take a look there in the meantime. It was a pretty small change..
2. I am new to kuma so not familiar with the implementation. But I'm curious what the pros and cons are of using `<Box as="div" {...props} className="..." />` instead of just `<div {...props} className="..." />` in the generated code. In the interests of keeping things as slim and simple as possible, I'd probably prefer not to use a wrapper component if possible. So I'd be interested to try out a pull request that switches to the latter, maybe based on a configuration variable somewhere. What do you think?

poteboy:
@mmkal 

For the first question, I have asked other maintainers to review this PR just in case, so please hang tight for a bit 🙏 In the meantime, I will definitely check your draft PR. I agree, this is a must-have feature.

For the second question, the use of `<Box>` is actually redundant. Historically, the `styled` API internally used `Box` (as they used to accept styled props), but I have decided not to anymore in order to align with other major CSS-in-JS libraries. So, nice catch! Please go ahead and replace it with a raw HTML tag. That way, the `styled` API is going to be subject to tree shaking.

mmkal:
Ok, if you're on board with all three changes, I could just combine

mmkal:
Update: I merged in support for `` styled(OtherComponent)` ... ` `` but not the `<Box as="div" />` -> `<div />` change as that was a little more invasive. I tried it but it left orphaned `import { Box } from '@kuma-ui/core'` statements, and removing those caused errors. It's non-essential for the project I'd like to try kuma on, so I thought best to leave it until later/let someone with a little more familiarity with this codebase try it.

So, this PR is ready for review! I updated the changeset and the PR title to reflect the broader scope.

Note: since there are quite a few commits in this branch now, I would suggest using squash-and-merge.

poteboy:
>  I tried it but it left orphaned import { Box } from '@kuma-ui/core' statements, and removing those caused errors.

This is fine as it is! If you are using a bundler like Vite or webpack, it will automatically tree-shake the unused import statements for you! Of course, it's up to you though.

mmkal:
> >  I tried it but it left orphaned import { Box } from '@kuma-ui/core' statements, and removing those caused errors.
> 
> 
> 
> This is fine as it is! With a bundler like Vite or Webpack, it will automatically tree-shake the unused import statements for you! Of course, it's up to you though.

Yeah that's true, but still it seemed like something worth cleaning up. If I get the time next week maybe I'll give it another try after this is in.

poteboy:
Looks good to me! Can you guys take a look at this PR? @kotarella1110 @taishinaritomi

kotarella1110 (approved):
Awesome work!

PR #343, opened 2023-09-24
Implement ForwardRef in Components

### LINKED ISSUE #256: Implementation of ForwardRef needed for proper handling of refs

poteboy:
Fix #256 
This PR introduces the use of `forwardRef` in our components.

## Attribution
The `forwardRef` implementation is based on Chakra UI's solution found at:
[Chakra UI forwardRef Implementation](https://github.com/chakra-ui/chakra-ui/blob/main/packages/core/system/src/forward-ref.tsx)

### DIFF
--- a/packages/core/src/components/forwardRef.ts
+++ b/packages/core/src/components/forwardRef.ts
@@ -1,0 +1,44 @@
+/**
+ * This code is based on the implementation from Chakra UI, an open-source UI library.
+ * The original implementation can be found at:
+ * https://github.com/chakra-ui/chakra-ui/blob/main/packages/core/system/src/forward-ref.tsx
+ *
+ * MIT License
+ * Copyright (c) 2019 Segun Adebayo
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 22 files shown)


packages/core/src/components/forwardRef.ts
```diff
@@ -0,0 +1,44 @@
...
+    RightJoinProps<PropsOf<Component>, Props> & {
+      as?: As;
+    }
+  >
+): ComponentWithAs<Component, Props> {
+  return forwardReactRef(component) as unknown as ComponentWithAs<
+    Component,
+    Props
+  >;
+}
```
kotarella1110, line 44: good 👍

kotarella1110 (approved):
LGTM 🚀

yukukotani (approved):
When using `as` prop, the type of `ref` prop is an intersection of original one and forwarded one.

image


This will cause type error with the code below since `buttonRef` is not match to `RefObject<HTMLDivElement>`.

```tsx
const buttonRef = useRef<HTMLButtonElement>(null);
<Box as="button" ref={buttonRef}></Box>
```

But I think we can merge this PR and tackle this issue as a followup because this issue is not only in `ref` but also in other props such as `onClick`.

poteboy:
@yukukotani 
Good point. Let's merge this PR for now and address this issue in the following PR!

PR #347, opened 2023-09-27
check 'styled' import binding before transforming

### LINKED ISSUE #346: Don't transform styled-components
Hi again! Similar to #341 - we're looking into migrating a styled-components project to kuma, but don't want to jump over fully yet. I noticed that the kuma webpack plugin (via the nextjs plugin) is transforming _all_ `` styled.xxx` ` `` code into kuma-ish code. We want to be deliberate and safe about this, though, especially since styled-components syntax parity is so new(!).

So, would it be possible to check that the `styled` import actually came via `import {styled} from '@kuma-ui/core'` before transforming it?

It should be possible to mix-and-match in a single file, so for a file like this:

```tsx
import styled from 'styled-components'
import {styled as kuma} from '@kuma-ui/core'

export const One = styled.div`
  color: red;
`
export const Two = kuma.div`
  color: blue;
`
```

`One` should be left alone and `Two` should be transformed.

I'd be happy to submit another PR, if you could point me in the right direction again (I'm not sure how `bindings` are initialized, or where would be the right place to check where they're imported from).

mmkal:
Fixes #346

### DIFF
--- a/.changeset/wet-rabbits-kick.md
+++ b/.changeset/wet-rabbits-kick.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/compiler": patch
+---
+
+check 'styled' import binding before transforming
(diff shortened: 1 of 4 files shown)


poteboy (approved):
LGTM 🚀

PR #348, opened 2023-09-27
Object type determination with strict equivalence operator

nojiritakeshi:
Sorry it's a really small correction, but I made the correction.

### DIFF
--- a/.changeset/green-gifts-deliver.md
+++ b/.changeset/green-gifts-deliver.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/core": major
+---
+
+fixed the type checking operator.
--- a/.changeset/green-gifts-deliver.md
+++ b/.changeset/green-gifts-deliver.md
@@ -1,5 +1,5 @@
 ---
-"@kuma-ui/core": major
+"@kuma-ui/core": patch
 ---
 
 fixed the type checking operator.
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/object.ts
```diff
@@ -29,7 +29,7 @@ export function flattenObject<const T, T2 extends NestedObject<T>>(
     if (!Object.hasOwn(object, key)) continue;
     const value = object[key];
 
-    if (typeof value == "object" && value !== null) {
+    if (typeof value === "object" && value !== null) {
       const _object = flattenObject(value as NestedObject);
       for (const _key in _object) {
         if (!Object.hasOwn(_object, _key)) continue;
```
nojiritakeshi, line 35: My question is, is this process necessary?
If it seems unnecessary, I will remove it and commit.

.changeset/green-gifts-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kuma-ui/core": major
```
poteboy, line 2:
```suggestion
"@kuma-ui/core": patch
```

packages/core/src/utils/object.ts
poteboy: @taishinaritomi 
IIRC I believe @taishinaritomi  made this change, can you confirm?

taishinaritomi: It ensures that we're only processing the object's own properties and not the properties from its prototype chain.

スクリーンショット 2023-09-28 10 33 17

スクリーンショット 2023-09-28 10 33 28

nojiritakeshi: @taishinaritomi 
I see.... .very informative. thank you...!

poteboy (approved):
LGTM!

Issue #349, opened 2023-10-01
Custom theme does not work correctly with pseudo-elements


tomoyuki-hiranuma:
# Description
Custom theme does not work correctly with pseudo elements, such as `_hover`, `_active` in nextjs project, as the title says.

## Environment
Nextjs@13.4.19
kuma-ui@1.4.1
pnpm@8.7.4
node@18.17.1

## Demo
1. Pull [kumaui-next-problem-minimal-sample](https://github.com/tomoyuki-hiranuma/kumaui-next-problem-minimal-sample)
2. Run `pnpm install`
3. Run `pnpm run dev`
4. Open [http://localhost:3000](http://localhost:3000)
5. You can see the `bgColor` in the text does not changed when hovering state, and the `bgColor` in the button does not changed when active state.

poteboy:
@tomoyuki-hiranuma 
Thank you for raising this issue. The generation of styles is handled in `packages/system/src/generator.ts`, so that area seems to be suspicious.

tomoyuki-hiranuma:
@poteboy 
Thank you for helpful information.
I will try it 🙌

PR #352, opened 2023-10-02
Fix generator to work custom style in pseudo css correctly

tomoyuki-hiranuma:
This is for https://github.com/kuma-ui/kuma-ui/issues/349.

Mapping of custom styles in pseudo-elements when generating styles was not correct.
I couldn't setting custom style in test code in `src/system/generator.test.ts`, so I added `k.button` element in the next.js example as a test.


poteboy:
fix #349

taishinaritomi (approved):

PR #357, opened 2023-10-14
Enable `forwardRef` Usage in Components Created through the `styled` API

poteboy:
###  Summary
This PR allows Kuma UI's styled components created through the `styled` API to accept refs without triggering warnings. The unique alias assures that user-defined variable names are not in conflict with Kuma's internal usage of React.

### Changes
- Adjusted [`ensureReactImport`](https://github.com/kuma-ui/kuma-ui/blob/main/packages/babel-plugin/src/ensureReactImport.ts) to import React as `__React_Kuma_`.
- Modified [`replaceTaggedTemplate`](https://github.com/kuma-ui/kuma-ui/blob/main/packages/compiler/src/processTaggedTemplateExpression.ts) to use `__React_Kuma_.forwardRef` enabling the passing of refs.

### DIFF
--- a/packages/babel-plugin/src/ensureReactImport.ts
+++ b/packages/babel-plugin/src/ensureReactImport.ts
@@ -2,11 +2,13 @@ import { NodePath } from "@babel/core";
 import { types } from "@babel/core";
 
 /**
- * Ensures that a React import declaration exists in the given program.
- * If not, it adds the import declaration at the beginning of the program.
+ * Ensures that React is imported in the given program to satisfy JSX's Classic runtime requirement.
+ * This function adds a React import declaration at the beginning of the program with a unique alias to avoid potential naming conflicts.
  *
- * @param {NodePath<t.Program>} programPath - The path to the Program node in the AST.
- * @param {typeof t} t - The Babel types.
+ * Note: The unique alias "__React_Kuma_" is utilized in Kuma's compiler to safely refer to React (e.g., when using React.forwardRef to wrap styled components) without interfering with user-defined variable names.
+ *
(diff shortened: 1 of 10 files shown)


packages/babel-plugin/src/ensureReactImport.ts
```diff
@@ -2,29 +2,21 @@ import { NodePath } from "@babel/core";
...
-      [t.importDefaultSpecifier(t.identifier("React"))],
-      t.stringLiteral("react"),
-    );
-    programPath.unshiftContainer("body", reactImportDeclaration);
-  }
+  const reactImportDeclaration = t.importDeclaration(
+    [t.importDefaultSpecifier(t.identifier("__React_Kuma_"))],
+    t.stringLiteral("react"),
+  );
+  programPath.unshiftContainer("body", reactImportDeclaration);
```
yukukotani, line 21: 📝 (Just a memo of my understanding) We already ensure that React is imported here, but we can't determine the specifier name if default import exists, therefore we need to insert another import for Kuma.

yukukotani (approved):
LGTM!

yukukotani:
[IMO] Naming it `__KUMA_REACT__` might be better to make it consistent with  `__KUMA_USER_THEME__`.

packages/babel-plugin/src/ensureReactImport.ts
poteboy: Thanks for the clarification. I've incorporated your comment into the JSDoc.

poteboy:
> [IMO] Naming it __KUMA_REACT__ might be better to make it consistent with __KUMA_USER_THEME__.

Right, fixed it 👍

PR #364, opened 2023-10-21
[Experiment] Sonik support

yukukotani:
https://github.com/sonikjs/sonik


packages/vite/src/index.ts
```diff
@@ -72,6 +72,7 @@ export default function kumaUI(): Plugin {
       sheet.reset();
       return (
         `import "${virtualModuleId}/${cssRelativePath}";
+        const __KUMA_CSS__ = \`${css}\`;
```
yukukotani, line 75: :memo: Sonik (Hono) doesn't support importing CSS as ES Modules on server side, so injecting raw CSS here

example/sonik/app/routes/_layout.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <small>© 2023 your name</small>
+        </footer>
+      </div>
+    </body>
+  );
+
+  const registry = createStyleRegistry();
+  renderToString(<StyleRegistry registry={registry}>{body}</StyleRegistry>); // Render here to determine the initial CSS.
+  const runtimeCss = registry.styles();
+  registry.flush();
```
yukukotani, line 30: :memo: Sonik doesn't have a feature that equivalent to `useServerInsertedHTML` so render twice like in next.js Pages Router

PR #368, opened 2023-10-22
fix(vite): fail to load components in the directory above package root

### LINKED ISSUE #344: Vite plugin does not work correctly when components imported from another package exists.
## Description
As the title says, **in the project using vite plugin(`@kuma-ui/vite`)**, we cannot import from another package's component.

### Situation
Under the monorepo environment (such as pnpm-workspace, turborepo),
we often create a package named 'ui' or like that and put UI components into it.
But, some other frameworks like Next.js can treat the component imported from another package, the vite plugin cannot.

<br />

## Demo
1. Pull [AsPulse/minimal-sample-kumaui-vite-problem
](https://github.com/AsPulse/minimal-sample-kumaui-vite-problem)
1. `pnpm install`
1. `pnpm -F vite-project dev`
1. Open [localhost:3000](http://localhost:3000) in your browser
1. See that the errors described above appear in both the browser console and terminal.

### Expected behavior
The vite plugin load [ui/components/redText.tsx](https://github.com/AsPulse/minimal-sample-kumaui-vite-problem/blob/main/packages/ui/components/redText.tsx) on loading [vite-project/src/App.tsx](https://github.com/AsPulse/minimal-sample-kumaui-vite-problem/blob/main/packages/vite-project/src/App.tsx), the red text appears.

### Actual behavior
The error appears in the browser console:
```
Failed to load resource: the server responded with a status of 404 (Not Found)
redText-2626507544.css:1 
```
terminal:
```
Failed to load url ui/components/redText-2626507544.css (resolved id: ui/components/redText-2626507544.css) in 
(project root path)/ui/components/redText.tsx. Does the file exist?
```

AsPulse:
- fix #344 

### Current behavior
For example, when loading `./src/foo.tsx`, an import sentence is inserted like below:
```ts
import "virtual:kuma-ui/src/foo-(filehash).css";
```
(this logic is written in `transform()`)
https://github.com/kuma-ui/kuma-ui/blob/0829b750ba1cf7eaef4206f6798c0394f01abceb/packages/vite/src/index.ts#L68-L76

However, using the relative path to their files for the import path is harmful when loading components in the directory above the current directory.

This situation often occurs in monorepo projects.
Let's say we have a project which has two packages `ui` and `website`.

When importing `packages/ui/foo.tsx` from whatever component exists in `website`, the import sentence is the following:
```ts
import "virtual:kuma-ui/../ui/foo-(filehash).css";
```
(**Notice that Vite forces the current directory to be package root,** so the current directory is set to `packages/website`.)
This is taken as `ui/foo-(filehash).css`, which can't recognize because it doesn't start with `virtual:kuma-ui`.

<br />

### Behavior after this PR
The id which isn't related to the relative path is used for the import path, which is generated like this:
```ts
const cssTableId = `${Date.now().toString(16)}x${cssTable.length}`;
const url = `${virtualModuleId}/${cssTableId}.css`;
```

### DIFF
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -38,7 +38,10 @@ export default function kumaUI(): Plugin {
     }
   }
 
-  const cssLookup: { [key: string]: string } = {};
+  const cssTable: {
+    url: string,
+    css: string,
+  }[] = [];
   const virtualModuleId = "virtual:kuma-ui";
 
   const userTheme = theme.getUserTheme();
@@ -62,23 +65,21 @@ export default function kumaUI(): Plugin {
       const result = transform(code, id);
       if (!result?.code) return;
       const css = (result.metadata as unknown as { css: string }).css || "";
-      const cssFilename = path.normalize(
-        `${id.replace(/\.[jt]sx?$/, "")}-${generateHash(css)}.css`,
-      );
-      const cssRelativePath = path
-        .relative(process.cwd(), cssFilename)
-        .replace(/\\/g, path.posix.sep);
-      cssLookup[cssRelativePath] = css;
+
+      const cssTableId = `${Date.now().toString(16)}x${cssTable.length}`;
+      const url = `${virtualModuleId}/${cssTableId}.css`;
+      cssTable.push({ url, css });
+
       sheet.reset();
       return (
-        `import "${virtualModuleId}/${cssRelativePath}";
+        `import "${url}";
 ` + result.code
       );
     },
     load(url) {
       if (!url.startsWith(`\0${virtualModuleId}`)) return undefined;
-      const id = url.slice(`\0${virtualModuleId}`.length + 1);
-      return cssLookup[id];
+      const urlContent = url.slice(`\0`.length);
+      return cssTable.find(c => c.url === urlContent)?.css ?? undefined;
     },
     resolveId(importeeUrl) {
       if (!importeeUrl.startsWith(virtualModuleId)) return undefined;
@@ -86,6 +87,7 @@ export default function kumaUI(): Plugin {
     },
     handleHotUpdate({ server }) {
       sheet.reset();
+      cssTable.length = 0;
       server.ws.send({ type: "full-reload" });
     },
     configResolved(config) {


packages/vite/src/index.ts
```diff
@@ -62,30 +65,29 @@ export default function kumaUI(): Plugin {
...
       const css = (result.metadata as unknown as { css: string }).css || "";
-      const cssFilename = path.normalize(
-        `${id.replace(/\.[jt]sx?$/, "")}-${generateHash(css)}.css`,
-      );
-      const cssRelativePath = path
-        .relative(process.cwd(), cssFilename)
-        .replace(/\\/g, path.posix.sep);
-      cssLookup[cssRelativePath] = css;
+
+      const cssTableId = `${Date.now().toString(16)}x${cssTable.length}`;
```
kotarella1110, line 69: @AsPulse 
Thank you for your PR!
It seems that the import URL for the virtual module changes every time we build. Is it possible to hash the relative (or absolute) paths in the CSS to ensure that they remain unchanged even after building?

AsPulse: Of course it is possible,
but since, when the component changed, the import path for the virtual module were still changing every time we build.
It is because current behavior adds hashed the "content" of the CSS to path.

If it's not too much trouble, could you let me know why keeping import path is better...?

kotarella1110: I believe that with your current code, the hash is being changed every time a file is saved, even if the Kuma UI component styles remain unchanged. This leads to a change in the CSS import path each time. I'm commenting with the intention of requesting that the hash be changed only when there is a change in the styles, just like it was before.

By doing this, we can avoid the scenario where CSS is reloaded with HMR every time a file is saved, even when there is no change in the styles. If we only change the hash when there is a change in the styles, HMR should work smoothly.

ref: https://discord.com/channels/1119672933477011598/1127180003872886794/1174362017491058718

kotarella1110: Additional information: it seems that a full reload is triggered in the Vite plugin's handleHotUpdate, so even when saving a file without any code changes, all CSS and JS are being loaded every time. The handleHotUpdate here is unnecessary, so please remove it.

AsPulse:
Based on the above suggestion, we have changed the behavior to use such path names. (thanks @kotarella1110 ;)
```ts
import "virtual:kuma-ui/(hash of path)/(filename)-(hash of css content).css";
```

e.g.) 
```ts
import "virtual:kuma-ui/316724801/App-708560247.css";
```


With this change, CSS paths that have not been modified are now maintained and do not need to be reloaded when HMR.

kotarella1110 (approved):
Awesome fixed 🎊
Thank you!

kotarella1110 (approved):

poteboy (approved):
Thank you so much 🎊

Issue #353, opened 2023-10-02
Cannot use background-image CSS property.


yodakaEngineer:
# Description
I wrote this code.
```
<Box
      backgroundImage={'url(/topImage.jpg)'}
      width={'100%'}
      height={'300px'}
      backgroundSize={'cover'}
    >
      <div ref={ref}>
      </div>
    </Box>
```

But, background-image was not generated. Console has no error.
<img width="246" alt="image" src="https://github.com/kuma-ui/kuma-ui/assets/44350989/d9aa186c-9b72-4548-ac2a-7b63ac119e10">

# Environment
Nextjs@13.4.4
kuma-ui@1.5.0
kuma-ui/next-plugin@1.1.10
node@16.20.0

poteboy:
Fixed in #371

PR #371, opened 2023-10-29
fix: background-image property unable to use `url()`

Asuka109:
Refer to: https://github.com/kuma-ui/kuma-ui/issues/353

The style generator incorrectly matches expressions like `url(./image.png)` as custom styles. Because custom styles cannot be used within the CSS value function, I have added additional conditional checks for this case.
However, I believe this is a workaround and perhaps we should consider refactoring the logic in L37?

### DIFF
--- a/.changeset/old-buckets-laugh.md
+++ b/.changeset/old-buckets-laugh.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/system": patch
+---
+
+fix: background-image property unable to use url
--- a/packages/core/src/themeTokens.ts
+++ b/packages/core/src/themeTokens.ts
@@ -14,10 +14,11 @@ export type NumberToken =
   | "letterSpacings"
   | "breakpoints";
 
-export type InputThemeTokens = Partial<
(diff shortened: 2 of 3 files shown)


poteboy:
Ah, I see. Thanks for clarifying that for me.
By the way, what do you think would be the ideal solution at its core?

Asuka109:
@poteboy Introducing postcss might be a good choice. Transforming CSS operations from string-based to AST-based should make kuma more robust.

poteboy:
@Asuka109 
PostCSS sounds like a good idea. Thanks for the suggestion. I'll give it some thought.
For the current scope of the issue, this PR seems like a good approach and I'm thinking of merging it. 
Just one question: What's the reason for the changes to the InputThemeTokens type?

Asuka109:
@poteboy I added it while reading the source code. It doesn't have any purpose, just makes it easier for me to read. Do I need to revert this commit?

poteboy (approved):
> I added it while reading the source code. It doesn't have any purpose, just makes it easier for me to read. Do I need to revert this commit?

No worries, I agree this version is easier to read. LGTM!

Issue #288, opened 2023-08-16
Consider Replacing `ts-pattern` with TypeScript's satisfies or ESLint's `switch-exhaustiveness-check` for Performance Gains 🤔


poteboy:
Currently, in Kuma UI, we utilize the [`ts-pattern`](https://github.com/gvergnaud/ts-pattern) for exhaustive type inference checks through pattern matching, as demonstrated below:

```ts
const isComponentProps =
  (componentName: ComponentName) => (propName: string) => {
    return match(componentName)
      .with("Box", () => isBoxProps(propName))
      // other code
      .with("Grid", () => isGridProps(propName))
      .exhaustive();
  };
```

However, having `ts-pattern` as a dependency in both `@kuma-ui/core` and `@kuma-ui/compiler` has raised some concerns:

- **Bundle Size** : We'd ideally like to minimize our bundle size, and avoiding this dependency can help achieve that.
- **Performance**: I conducted a benchmark comparison between the native JavaScript switch and ts-pattern. Unfortunately, ts-pattern seems to be considerably slower. Check out the code and results here:

```ts
import { match } from "ts-pattern";
import { add, cycle, complete, suite } from "benny";

export type Data =
  | { type: "text"; content: string }
  | { type: "number"; value: number };

export const testData: Data[] = Array.from({ length: 1000 }, (_, index) =>
  index % 2 === 0
    ? { type: "text", content: "Hello, world!" + index }
    : { type: "number", value: index }
);

function process(data: Data) {
  return match(data)
    .with({ type: "text" }, (x) => x.content.length)
    .with({ type: "number" }, (x) => x.value * 2)
    .exhaustive();
}

function processNative(data: Data) {
  switch (data.type) {
    case "text":
      return data.content.length;
    case "number":
      return data.value * 2;
  }
}

suite(
  "Pattern Matching Benchmark",

  add("ts-pattern", () => {
    for (const data of testData) {
      process(data);
    }
  }),

  add("Native JS", () => {
    for (const data of testData) {
      processNative(data);
    }
  }),

  cycle(),

  complete()
);
```

Result:

```
ts-pattern: 5,985 ops/s, ±0.51%     | slowest, 98.31% slower
Native JS: 354,529 ops/s, ±0.28%   | fastest
```
- **HMR Speed**: For Hot Module Replacement performance improvements, removing runtime intervention from ts-pattern might be beneficial.

[...]

PR #372, opened 2023-10-31
chore: refactor ts-pattern to use switch/if-else

AdePhil:
Fixes https://github.com/kuma-ui/kuma-ui/issues/288

### DIFF
--- a/.changeset/two-geckos-check.md
+++ b/.changeset/two-geckos-check.md
@@ -1,0 +1,7 @@
+---
+"@kuma-ui/babel-plugin": patch
+"@kuma-ui/compiler": patch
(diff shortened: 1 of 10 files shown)


poteboy:
@AdePhil 
Thank you for the  PR! I'll review it anytime soon 👍

yukukotani (approved):

poteboy (approved):
Thanks for the PR! In the `componentList.ts`, you've used a switch case with satisfies never, allowing us to maintain exhaustiveness when something new is added to the case, without using `ts-pattern`. Additionally, you've removed the `ts-pattern` dependencies from packages.json. Overall, this PR is perfect! Thank you!

PR #380, opened 2023-11-23
avoid false-positive hasDynamicProps

yukukotani:
Fix `Error: StyleSheetRegistry: id: `` not found in idIndexesMap.` error.

This error occurs on unmounting the code below. This is because Kuma renders DynamicBox despite the generated CSS will be empty.

```tsx
const isHoge = false;
<Box color={isHoge ? "red" : undefined} />;
```

This PR fixes `hasDynamicProps` to be more strict.

### DIFF
--- a/.changeset/tiny-pots-sparkle.md
+++ b/.changeset/tiny-pots-sparkle.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/core": patch
+---
+
+avoid false-positive hasDynamicProps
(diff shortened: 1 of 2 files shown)


poteboy (approved):
@yukukotani 
This modification seems valid to me. However, do you think `StyleSheetRegistry` should throw an error when the id is not found in `idIndexesMap` in the first place? With this PR, those errors might not occur, but I'm not sure. I feel like this operation (`StyleSheetRegistry.remove`) completes just fine without throwing an error

https://github.com/kuma-ui/kuma-ui/blob/main/packages/core/src/registry/StyleSheetRegistry.ts#L42

yukukotani:
@poteboy It sounds reasonable to just do nothing in `StyleSheetRegistry .remove` instead of throwing an error. In addition, It's better if `StyleSheetRegistry.add`  ignores empty className, and `StyleGenerator` explicitly returns null instead of empty className. Will work on those refactors as a following PR.

PR #383, opened 2023-11-24
chore(vite): specify vite version as a peer-dependency

AsPulse:
As the title says.
Vite 5 was recently released and is working fine on my end.
If there are no problems, I would be willing to include Vite 5 in the support.


yukukotani (approved):
Thanks!

PR #385, opened 2023-11-29
refactor: improve toCssUnit logic to reduce bundle size

kotarella1110:


packages/system/src/toCSS.ts
```diff
@@ -1,4 +1,61 @@
-export const toCssUnit = (token: string | number) => {
-  if (typeof token === "string") return token;
-  else return `${token}px`;
+/**
+ * An object of css properties that don't accept values with units
+ * @link https://github.com/emotion-js/emotion/tree/main/packages/unitless
+ */
+const unitless = {
```
yukukotani, line 5: I know this is from emotion, but why is this an object instead of a list of keys?

packages/system/src/toCSS.ts
```diff
@@ -1,4 +1,61 @@
...
+  stopOpacity: 1,
+  strokeDasharray: 1,
+  strokeDashoffset: 1,
+  strokeMiterlimit: 1,
+  strokeOpacity: 1,
+  strokeWidth: 1,
 };
+
+export function toCssUnit(propertyName: string, value: string | number) {
+  if (typeof value === "number") {
```
yukukotani, line 57: (Maybe off-topic from this PR) Do you think string value and number value should be treated as different? Users may not differentiate `width="1" and `width={1}`, I think.

packages/system/src/toCSS.ts
poteboy: O(1) lookup?

yukukotani: Make sense!

poteboy (approved):
LGTM!

yukukotani (approved):

packages/system/src/toCSS.ts
```diff
@@ -1,4 +1,61 @@
...
+  floodOpacity: 1,
+  stopOpacity: 1,
+  strokeDasharray: 1,
+  strokeDashoffset: 1,
+  strokeMiterlimit: 1,
+  strokeOpacity: 1,
+  strokeWidth: 1,
 };
+
+export function toCssUnit(propertyName: string, value: string | number) {
```
kotarella1110, line 56: Note: In emotion, the implementation seems to be as follows
https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/serialize/src/index.js#L31-L60

packages/system/src/toCSS.ts
kotarella1110: Yes, I think it is better to treat them as different, along with popular libraries such as Emotion.

https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/serialize/src/index.js#L31-L60

Issue #376, opened 2023-11-05
`_hover` only applies the first key if it uses theme tokens


Nedomas:
I'm encountering a bug where `_hover` functionality breaks if any of the keys use theme tokens.

```ts
// kuma.config.ts
import { createTheme } from '@kuma-ui/core'

const theme = createTheme({
  colors: {
    primary: 'blue',
    secondary: 'green',
  },
})

// Component.tsx
import { Button } from '@kuma-ui/core'

// the BUG: only color: 'green' will be applied on hover, but not fontSize
<Button
  color='colors.primary'
  _hover={{
    color: 'colors.secondary',
    fontSize: '100px',
  }}
>
  Text
</Button>

// no BUG: both color: 'red' and fontSize: '100px' will be applied on hover since no design tokens are used
<Button
  color='colors.primary'
  _hover={{
    color: 'red',
    fontSize: '100px',
  }}
>
  Text
</Button>
```

By the way, thank you for a great library!

poteboy:
@Nedomas 
Thank you for raising the issue. I just tested this in my local environment, and it indeed appears to be a bug, seemingly related to this one: #352.

P.S. Thanks for the compliment! 🫶

poteboy:
As mentioned in [PR #381](https://github.com/kuma-ui/kuma-ui/pull/381), it appears that pseudo props with array and theme tokens are not functioning correctly. I will address this issue while working on this particular problem as well.

PR #390, opened 2023-12-13
fix:apply theme tokens to pseudo-class

nojiritakeshi:
fix: https://github.com/kuma-ui/kuma-ui/issues/376
The original process would have overwritten all objects if the theme token had been applied to the pseudo-class.
Changed to add properties to objects.

### DIFF
--- a/packages/system/src/generator.ts
+++ b/packages/system/src/generator.ts
@@ -143,7 +143,7 @@ export class StyleGenerator {
 
     let css = `.${this.className} { ${this.style.base} }`;
     for (const [breakpoint, cssValue] of Object.entries(
-      this.style.responsive
+      this.style.responsive,
     )) {
       css += `@media (min-width: ${breakpoint}) { .${this.className} { ${cssValue} } }`;
     }
--- a/packages/system/src/generator.ts
+++ b/packages/system/src/generator.ts
@@ -50,6 +50,15 @@ export class StyleGenerator {
         !/^\w+\(.*\)$/.test(propValue) // exclude CSS functions
       );
     };
+
+    /**
+     * If the argument value is a user theme defined in the `kuma.config.ts` file, it is converted,
+     * otherwise the value of value is returned as is.
+     * @example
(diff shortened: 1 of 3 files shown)


packages/system/src/generator.ts
```diff
@@ -50,40 +50,46 @@ export class StyleGenerator {
         !/^\w+\(.*\)$/.test(propValue) // exclude CSS functions
       );
     };
-
-    for (const [propName, propValue] of Object.entries(props)) {
+    /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-return -- FIXME */
+    const convertStyle = (name: string, value: any) => {
```
poteboy, line 54: Can you add JSDoc and an @example tag for this function? Since the return type is any, it would at least make it easier for future readers to understand what the expected return value is. 🙏

packages/system/src/generator.ts
```diff
@@ -50,40 +50,46 @@ export class StyleGenerator {
...
-          styledProps[propName] = customStyle;
+          return customStyle;
         }
-      } else if (isStyledProp(propName)) {
-        styledProps[propName] = propValue;
-      } else if (isPseudoProps(propName)) {
+      } else if (isStyledProp(name)) {
+        return value;
+      }
+      return value;
```
poteboy, line 75: Like I have commented out examples of return values in other lines (for example: L55 & L66), it would be helpful if you could add a comment here as well indicating what kind of values can be returned

nojiritakeshi:
@poteboy 
I added a comment. What do you think?

poteboy:
@nojiritakeshi 
The added comments seem valid, thanks! However, it looks like the CI is failing, probably because the code isn't linted. Could you fix this?

nojiritakeshi:
@poteboy 
I also fixed the lint.
Sorry for not fixing it all at once!

poteboy (approved):
Thanks! Looks good to me!

PR #337, opened 2023-09-16
refactor(next): Liberation from the curse of '.next/cache/kuma/*.css'

taishinaritomi:
Needs to be verified.

PR #392, opened 2024-01-03
docs(install): remove unnecessary argument

MH4GF:
`KumaUIWebpackPlugin` currently does not accept arguments, so the documentation has been modified accordingly.

- ref: https://github.com/kuma-ui/kuma-ui/pull/337


poteboy (approved):
Thank you!

PR #394, opened 2024-01-08
Add "w" shorthand for width property

yodakaEngineer:
I personally implemented this because it was inconvenient for me to do without it.
If you feel strongly that implementing this is a bad idea, feel free to ignore this PR.
But since there is a short hand on the margin, this may just be a forgotten implementation.

### DIFF
--- a/packages/system/src/consumeStyleProps.ts
+++ b/packages/system/src/consumeStyleProps.ts
@@ -42,8 +42,6 @@ export type StyledProps<T extends ThemeSystemType = ThemeSystemType> =
     MaskProps &
     ColumnProps<T> &
     BackgroundProps &
-    // eslint-disable-next-line @typescript-eslint/no-duplicate-type-constituents -- FIXME
-    EffectProps &
     FilterProps;
 
 export type StyleFunction = (props: StyledProps) => ResponsiveStyle;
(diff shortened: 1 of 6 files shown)


packages/system/src/consumeStyleProps.ts
```diff
@@ -42,8 +42,6 @@ export type StyledProps<T extends ThemeSystemType = ThemeSystemType> =
     MaskProps &
     ColumnProps<T> &
     BackgroundProps &
-    // eslint-disable-next-line @typescript-eslint/no-duplicate-type-constituents -- FIXME
-    EffectProps &
```
yodakaEngineer, line 46: EffectProps was duplicate on line41. So remove it.

packages/system/src/props/layout.ts
```diff
@@ -19,10 +24,16 @@ export type LayoutProps<T extends ThemeSystemType = ThemeSystemType> = Partial<
     CSSProperties<"float" | "clear"> &
     CSSProperties<"objectFit" | "objectPosition"> &
     CSSProperties<"resize"> &
-    CSSProperties<"verticalAlign">
+    CSSProperties<"verticalAlign"> & {
+      /**
+       * @see margin
```
poteboy, line 29:
```suggestion
       * @see width
```

packages/system/src/consumeStyleProps.ts
poteboy: thx :)

poteboy (commented):
Thanks for the PR!
Can you add tests to `consumeStyleProps.test.ts` to make sure that changes are valid?

poteboy (approved):
❤️

PR #396, opened 2024-01-10
Add shorthands for layout properties

yodakaEngineer:
I should have implemented this PR https://github.com/kuma-ui/kuma-ui/pull/394, but I forgot.
This is adding minW, maxW, h, minH, maxH shorthands.


poteboy (approved):
LGTM 🫶

PR #397, opened 2024-01-11
Fix next hmr

taishinaritomi:


poteboy:
The webpack plugin for Kuma UI was originally updating styles in the Dev environment by adding extracted CSS to SearchParams. The code looked like this:

```ts
const importCSS = `import ${JSON.stringify(
      `${this.utils.contextify(this.context, CSS_PATH)}?${params.toString()}`,
    )};`;
    callback(null, `${result.code}\n${importCSS};`);
```

However, after some [internal changes](https://github.com/vercel/next.js/discussions/59212) in Next.js from version 13.5 onward, styles changes in Client Components no longer reflect unless reloaded when using this method.

To address this, we now apply arbitrary changes to the actual CSS file set as Entry whenever there are style changes. This ensures that Next.js's HMR triggers appropriately.

poteboy (approved):

PR #398, opened 2024-01-12
Fix the error on invalid style

yukukotani:
The index of rules can be `-1` if the insertion fails.

https://github.com/kuma-ui/kuma-ui/blob/0013d4a17aaafa94e95e5adb404c6f3111e6563a/packages/core/src/registry/sheet/ClientStyleSheet.ts#L64-L76

Then, deleting them [here](https://github.com/kuma-ui/kuma-ui/blob/cb37a56310bb412801e96302c7f41176c068ae7c/packages/core/src/registry/StyleSheetRegistry.ts#L66) results in the error below.

```
DOMException: Failed to execute 'deleteRule' on 'CSSStyleSheet': The index provided (4294967295) is larger than the maximum index (0).
```

Fixing it by just ignoring negative index on deletion.

### DIFF
--- a/.changeset/eleven-papayas-try.md
+++ b/.changeset/eleven-papayas-try.md
@@ -1,0 +1,5 @@
+---
+"@kuma-ui/core": patch
+---
+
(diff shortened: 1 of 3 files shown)


poteboy (approved):
LGTM!

PR #410, opened 2024-05-05
Introduce a Rust Based Compiler

poteboy:
## Overview

This pull request introduces a rewritten version of the plugin originally implemented in Babel, now in Rust using the [Oxc](https://oxc-project.github.io/). This change is a first step towards v2 to migrate our AST manipulation from JavaScript to Rust, specifically targeting WebAssembly for improved performance.


poteboy:
Almost complete. Here are some remaining issues:

- In `ReplaceKWithBox`, when constructing the `JSXIdentifier` IR with `box_local_name`, there are issues with transferring ownership due to lifetime constraints, and currently, it's hardcoded.
- When deploying to Vercel, it fails with `failed to start cargo metadata`, so a custom script for Vercel might be necessary to build wasm.

packages/wasm/src/transform/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+     * with corresponding `Box` component. This allows usage of the `k` syntax as a shorthand for creating
+     * styled components, e.g. `<k.div>` is transformed to `<Box as="div">`.
+     */
+    fn visit_jsx_element(&mut self, elem: &mut JSXElement<'a>) {
+        if let Some(k_alias) = self.imports.get("k").cloned() {
+            if let Some(box_local_name) = self.imports.get("Box") {
+                let k_alias = k_alias.clone();
+                self.process_element(elem, &k_alias, &box_local_name.clone());
+            }
+        }
```
Boshen, line 243: https://github.com/oxc-project/oxc/blob/5081652bc10cdcfd10cf57bf0385ebc7eacd5769/crates/oxc_ast/src/visit/visit_mut.rs#L404

You need to call `walk_jsx_element_mut` otherwise it's not going to visit its children.

packages/wasm/src/transform/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+    pub fn process_element(
+        &mut self,
+        elem: &mut JSXElement<'a>,
+        k_alias: &str,
+        box_local_name: &str,
+    ) {
+        if let JSXElementName::MemberExpression(member_expr) = &elem.opening_element.name {
+            if let JSXMemberExpressionObject::Identifier(ident) = &member_expr.object {
+                let html_tag = member_expr.property.name.as_str();
+                let html_tag: &'a str = unsafe { std::mem::transmute(html_tag) }; // FIXME: Remove unsafe transmute
```
Boshen, line 192: https://github.com/oxc-project/oxc/blob/5081652bc10cdcfd10cf57bf0385ebc7eacd5769/crates/oxc_ast/src/ast_builder.rs#L64-L67

Maybe you want `new_atom` or `String::from_str_in`

taishinaritomi:
JSXIdentifier reference_id ?

```tsx
import { k } from \"@kuma-ui/core\";

export function Component() {
  const k = {}

  return <k.div />;
}
```

packages/wasm/src/transform/mod.rs
poteboy: @Boshen 
You are right, thank you so much for taking a look at the PR 🤝

taishinaritomi (approved):

PR #416, opened 2024-06-12
docs(website): remove references to non-existent options in install.mdx

apple-yagi:
## Overview
While reviewing the documentation for kuma-ui, I noticed that there was a reference to a non-existent option in the Vite plugin section. Therefore, I created a Pull Request to remove that reference.

https://github.com/kuma-ui/kuma-ui/blob/0bcf3b0c1dbcba43d4620219f9031d366365d3e6/packages/vite/src/index.ts#L9-L12


poteboy (approved):
@apple-yagi 
Thank you so much 🙌

PR #422, opened 2024-08-31
Compile `css` Tagged Template Literals in Wasm

poteboy:
We are in the process of migrating the entire Kuma UI compiler to Wasm. This PR is part of that effort.

In this PR, the processing of `css` tagged template literals, previously handled by [Babel](https://github.com/kuma-ui/kuma-ui/blob/main/packages/babel-plugin/src/processTaggedTemplateExpression.ts) and [ts-morph](https://github.com/kuma-ui/kuma-ui/blob/main/packages/compiler/src/processTaggedTemplateExpression.ts), has been migrated to Rust (Wasm). Although both `styled` and `css` need to be processed, this PR focuses on the `css` implementation.

### DIFF
--- a/packages/wasm/src/compile/visit/visit_tagged_template_expression.rs
+++ b/packages/wasm/src/compile/visit/visit_tagged_template_expression.rs
@@ -1,0 +1,110 @@
+use std::{
+    borrow::{Borrow, BorrowMut},
+    collections::HashMap,
+};
+
+use oxc_allocator::Allocator;
+use oxc_ast::{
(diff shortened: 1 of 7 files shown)


packages/wasm/src/compile/visit/visit_tagged_template_expression.rs
poteboy, line 1: How can I run tests on code that depends on JS FFI via wasm-bindgen without testing on the compiled Wasm file? Mocking externed JS functions is acceptable.

PR #438, opened 2025-11-26
🌎 Implement support for global styles

thomaslindstrom:
Allows styling outside of the Kuma class name scope.

```js
import { css } from "@kuma-ui/core";

export const globalStyles = css`
  :global(.dark-mode) & {
    color: #d0d7ff;
  }

  :global {
    ::view-transition(old) {
      opacity: 0;
    }
    ::view-transition(new) {
      opacity: 1;
    }
  }
`;
```

### DIFF
--- a/packages/sheet/src/sheet.test.ts
+++ b/packages/sheet/src/sheet.test.ts
@@ -154,9 +154,9 @@ describe("Sheet class", () => {
           color: blue;
         }
       `);
+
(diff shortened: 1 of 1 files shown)


poteboy (approved):
@thomaslindstrom 
Thanks for the PR, looks good! btw, do you want merge permissions?

